PR #29, opened 2026-05-01
fix(platform/linux/wlgrab): Swap R<->B for some DMA buffers

iankendrick:
## Summary
I was seeing Blue/Red swapped using my RX 6700. The issue didn't appear in the Web preview, nor did it occur outside of Headless mode. It happened regardless of VAAPI or SW encoder. I'm not a graphics programmer but I worked with Claude to isolate the issue to `wlgrab` and created a fix and tested it locally (colors are now correct on my Moonlight clients.)

It seems like the GBM format varies based on driver, GPU, and mesa version so it may not be a common issue.

## What changed, and why?

Essentially inspect the SHM format, if it is one that is known to use an RGBX format, then swap the Red and Blue when doing memcpy so encoder gets the correct colors (BGRX). More details from Claude:

```On little-endian systems, XBGR8888 and ABGR8888 DMA-buf frames arrive with
memory layout [R,G,B,X], but the encoder expects [B,G,R,X]. The previous
code copied all non-3bpp formats directly, producing tinted output for
compositors that report these formats. Add per-pixel R↔B swap for those
two fourcc values and unit tests covering single-pixel, multi-row, and
padded-stride cases.
```
## Testing

- [n/] I included screenshots or recordings for visible UI changes, when practical.
- [n/a] I updated docs or release notes when user-facing behavior changed.

## Notes
I'm unable to test on any systems using any other GPU or driver combination, so I can't confirm there aren't regressions.

### DIFF
--- a/src/platform/linux/portal_grab.cpp
+++ b/src/platform/linux/portal_grab.cpp
@@ -853,7 +853,7 @@ namespace portal {
           }
         }
       } else {
-        // 4bpp format (XRGB8888/ARGB8888) — direct copy
+        // 4bpp format: direct copy (XRGB8888/ARGB8888 are already [B,G,R,X] on LE)
         int copy_bytes = std::min((int) st.stride, img_out->row_pitch);
         for (int y = 0; y < copy_h; ++y) {
           std::memcpy(img_out->data + y * img_out->row_pitch,
--- a/src/platform/linux/portal_grab.cpp
+++ b/src/platform/linux/portal_grab.cpp
@@ -853,7 +853,7 @@ namespace portal {
           }
         }
       } else {
-        // 4bpp format: direct copy (XRGB8888/ARGB8888 are already [B,G,R,X] on LE)
+        // 4bpp format (XRGB8888/ARGB8888) — direct copy
         int copy_bytes = std::min((int) st.stride, img_out->row_pitch);
         for (int y = 0; y < copy_h; ++y) {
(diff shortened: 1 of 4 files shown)


papi-ux:
Hey there! Thanks a bunch for tracking this down and working on a fix. I just pulled it locally and did a cleanup pass:

What I verified so far:

  - Built the cleaned-up branch on Fedora/Bazzite.
  - `test_polaris` passes, including direct coverage for the `XBGR8888` / `ABGR8888` 4bpp swap behavior.
  - Smoke-tested a real Retroid Pocket 6 session through headless labwc with GPU-native capture disabled.

One note from the smoke test: my local labwc session used the 3bpp SHM path:

  `SHM screencopy buffer allocated: 1920x1080 stride=5760 format=875710274`
  `wlr: SHM screencopy is 3bpp BGR888; expanding to 4bpp BGRA`

I’m gonna carry this forward with the helper/test cleanup. Thanks again!

Cheers

papi-ux:
Thanks again for the first pass here. I pulled this direction into #30 and cleaned it up around the newer #31/#35 reports, including the reported 3bpp path and the four reported 4bpp SHM formats with unit coverage.

I’m going to keep #30 as the replacement PR so the final patch has the updated tests and writeup in one place.

iankendrick:
I'll go ahead and close this, thanks for picking up the fix and your work on this project!

PR #30, opened 2026-05-02
Refine wlgrab SHM channel swap handling

papi-ux:
## Summary

Carries forward the wlgrab SHM color-channel work from #29, updated with the latest #31/#35 reporter data.

- keep the existing 3bpp SHM conversion (`RGB` byte order -> `BGRA`)
- swap R/B for reported 4bpp SHM formats: `XRGB8888`, `ARGB8888`, `XBGR8888`, and `ABGR8888`
- keep unknown 4bpp SHM formats as copy-through
- keep the pixel-copy helpers covered by Linux unit tests

## Why

The latest #31 follow-up corrected the earlier theory: the regression is still in the true-headless SHM copy path, not the orphaned extcopy path. Reporters have now seen AMD/Mesa/headless labwc deliver 4bpp SHM frames where the byte order needs an R/B swap before Polaris hands the frame to the encoder.

This patch keeps the conversion narrow to the four reported 4bpp formats while preserving copy-through behavior for unknown formats.

## Validation

- `git diff --check`
- GitHub Actions passed on rebased head `d964709`
- Fedora 44 RPM artifact downloaded from Actions for local Retroid/Nova smoke testing
- Retroid Pocket 6 / Nova true-headless smoke test on `1.0.8.dd28c34` exercised the SHM path and colors looked correct

Remote Linux source-build testing was skipped for this update because the shared Linux checkout currently has unrelated tracked changes; using the sync-based remote helper would risk overwriting that work.

Co-authored-by: Ian Kendrick <iankendrick@gmail.com>

### DIFF
--- a/src/platform/linux/portal_grab.cpp
+++ b/src/platform/linux/portal_grab.cpp
@@ -853,7 +853,7 @@ namespace portal {
           }
         }
       } else {
-        // 4bpp format (XRGB8888/ARGB8888) — direct copy
+        // 4bpp format: direct copy (XRGB8888/ARGB8888 are already [B,G,R,X] on LE)
         int copy_bytes = std::min((int) st.stride, img_out->row_pitch);
         for (int y = 0; y < copy_h; ++y) {
           std::memcpy(img_out->data + y * img_out->row_pitch,
--- a/src/platform/linux/portal_grab.cpp
+++ b/src/platform/linux/portal_grab.cpp
@@ -853,7 +853,7 @@ namespace portal {
           }
         }
       } else {
-        // 4bpp format: direct copy (XRGB8888/ARGB8888 are already [B,G,R,X] on LE)
+        // 4bpp format (XRGB8888/ARGB8888) — direct copy
         int copy_bytes = std::min((int) st.stride, img_out->row_pitch);
         for (int y = 0; y < copy_h; ++y) {
           std::memcpy(img_out->data + y * img_out->row_pitch,
--- a/src/platform/linux/wlgrab.cpp
+++ b/src/platform/linux/wlgrab.cpp
@@ -3,7 +3,6 @@
  * @brief Definitions for wlgrab capture.
  */
 // standard includes
-#include <cstring>
 #include <thread>
 
 // local includes
@@ -16,6 +15,7 @@
 #include "src/video.h"
 #include "vaapi.h"
 #include "wayland.h"
+#include "wlgrab_pixel_copy.h"
 
 using namespace std::literals;
 
@@ -263,7 +263,6 @@ namespace wl {
         int copy_w = std::min((int) dmabuf.shm_info.width, img_out->width);
         int src_stride = dmabuf.shm_info.stride;
         int dst_stride = img_out->row_pitch;
-        int copy_bytes = std::min(src_stride, dst_stride);
         const bool is_3bpp = (src_stride == dmabuf.shm_info.width * 3);
 
         if (is_3bpp) {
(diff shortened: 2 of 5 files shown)


giftederic:
Adding one more data point from testing the refined direction here.

On my GMK EVO-X2 / AMD Ryzen AI Max+ 395 / Radeon 8060S system with Mesa 26.0.5, the affected true-headless SHM path reported:

```text
Using SHM screencopy (preferred for this runtime)
SHM screencopy buffer allocated: 1920x1080 stride=7680 format=875709016
wlr: capture_transport=shm frame_residency=cpu frame_format=bgra8
```

`875709016` is `0x34324258` / `DRM_FORMAT_XBGR8888`.

With current copy-through behavior for that 4bpp SHM format, Nova/Android showed the same blue/purple color symptom as #31. I applied a small local test patch that swaps R/B for `XBGR8888` / `ABGR8888` during the SHM copy path, and the same stream immediately looked correct.

So this AMD/Mesa stack appears to be a counterexample to treating all 4bpp `XBGR8888` SHM frames as already-safe copy-through. I agree the final fix probably should not blindly swap every `XBGR8888` frame, since another AMD/Mesa report says that regresses their colors, but this test suggests there is still a real 4bpp `XBGR8888` case that needs either a better runtime discriminator or an explicit guarded quirk.

Other context from this machine:

- The issue only matters on true headless SHM capture.
- When Polaris probes windowed labwc and gets DMA-BUF, colors are correct without this SHM path.
- VAAPI encode works; current good runtime is `capture_transport=dmabuf frame_residency=gpu frame_format=bgra8`.

Happy to test a narrower discriminator/quirk patch on this hardware.

profcelix:
> ## Summary
> 
> Carries forward the wlgrab SHM color-channel work from #29, but narrows it after #31 showed the 4bpp swap was too broad.
> 
> - keep the channel-changing conversion only for the confirmed 3bpp SHM path (`BGR888`/RGB byte order -> `BGRA`)
> - treat 4bpp SHM frames as conservative copy-through, including `XBGR8888`, `ABGR8888`, and `XRGB8888`
> - move SHM pixel-copy helpers into shared code used by production and tests
> - use `DRM_FORMAT_*` constants in Linux-only unit coverage
> 
> ## Why
> 
> The Retroid smoke test exercised the active 3bpp SHM path, where the stream looked correct. Issue #31 reports that the unconditional 4bpp `XBGR8888` / `ABGR8888` swap corrupts colors on RX 9070 XT + Mesa 26.0.6 because those frames are already in the expected byte order.
> 
> This version avoids using DRM fourcc alone as a runtime signal for red/blue swapping. If we need a hardware-specific 4bpp quirk later, it should be gated on a reliable runtime discriminator instead of applying globally.
> 
> ## Validation
> 
> - `git diff --check HEAD`
> - `~/bin/polaris-remote.sh cmake --build build -j8 --target test_polaris`
> - `~/bin/polaris-remote.sh ctest --test-dir build/tests --output-on-failure -R '^test_polaris$'`
> - Prior Retroid Pocket 6 smoke test through headless labwc with GPU-native capture disabled exercised the 3bpp SHM path (`BGR888 -> BGRA`) and looked correct
> 
> Co-authored-by: Ian Kendrick <iankendrick@gmail.com>

The 4bpp swap was the right direction.

I led this astray. My initial report pointed the 4bpp swap in PR #30, but my format is 875709016 which is DRM_FORMAT_XRGB8888 was never in the swap list to begin with, so PR #30 couldn't have caused my bug. 

Adding XRGB8888 and ARGB8888 alongside XBGR8888/ABGR8888 in the 4bpp SHM copy fixed my issue. All four formats would arrive in RGB byte order on AMD + Mesa + headless labwc. My updated extcopy theory was a detour, fixing SHM at the copy level aligns with where you were already headed in PR #29/#30. 

I apologize for the mix-up. I think the four-format 4bpp swap covers every AMD GPU that's reported this so far.

Issue #44, opened 2026-05-04
Track Linux headless dashboard preview backoff


papi-ux:
## Summary

Two tester reports point at the same Linux headless runtime surface:

1. CachyOS / RTX 5080 headless stream logs spam:

```text
display_preview: Failed to capture cage screenshot on socket wayland-1 output=HEADLESS-1
wlr: capture_transport=shm frame_residency=cpu frame_format=bgra8; capture will incur an extra CPU-side copy/conversion path
```

They also report low FPS in headless mode, with no problem when running without headless.

2. Another tester using Polaris for a Quest 3 / VR HMD target reports performance is too rough to compete with Sunshine, seeing about 60 FPS on an RTX 5070.

## Initial read

These are likely related but should be validated separately:

- The `display_preview` warning is the web dashboard preview path failing to grab a still frame from the private labwc compositor. That should not mean the Moonlight/Nova stream failed, but repeated polling can create noisy logs and may deserve a preview-disable or preview-backoff path.
- `capture_transport=shm frame_residency=cpu` is a real performance warning. Headless labwc can fall back to SHM/RAM capture, which adds CPU-side copies/conversion and can cap practical FPS, especially for high refresh clients such as VR HMDs.
- On NVIDIA, logs should be checked for `Build features: cuda=...` and `Attempting to use NVENC without CUDA support. Reverting back to GPU -> RAM -> GPU`, because that path can be much slower than Sunshine's GPU-native path.

## Details to request from testers

Ask for the full session startup log around:

- Polaris version and package type
- OS / desktop / driver version
- `Build features: cuda=...`
- `headless_mode`, `linux_use_cage_compositor`, `linux_prefer_gpu_native_capture`, `capture`, `encoder`
- requested client resolution / FPS / codec
- `Creating encoder [...]`
- `Using frame converter [...] target_device=... target_residency=... target_format=...`
- `wlr: capture_transport=... frame_residency=... frame_format=...`
- `session_pacing: policy=... target_fps=...`
- whether the Polaris web UI live preview page was open while streaming
- `command -v grim` output for preview capture availability

## Candidate fixes to investigate

- Add a setting or adaptive backoff to disable/silence live preview capture when cage screenshot capture repeatedly fails.
- Make preview capture dependency/status visible in diagnostics instead of spamming warning logs.
- Improve NVIDIA headless performance guidance and detection when the active path is SHM/CPU or NVENC-without-CUDA fallback.
[...]

PR #45, opened 2026-05-04
Improve headless performance diagnostics

papi-ux:
## Summary

- Rate-limit repeated cage preview capture failures so dashboard preview polling does not spam logs during otherwise healthy streams.
- Add shared capture-path diagnostics for SHM/CPU capture, CPU encoder upload, GPU-native capture, and CUDA-disabled NVENC fallback.
- Document the Linux headless performance markers testers should include, and add `grim` as a Linux runtime/package dependency for dashboard preview capture.

## Validation

- `git diff --check`
- `g++ -std=c++23 -I. -Ithird-party/googletest/googletest/include -fsyntax-only tests/unit/test_stream_stats.cpp` was attempted, but this local environment is missing Boost headers (`boost/core/noncopyable.hpp`).

## Notes

This targets #44. The stream path and dashboard preview path remain separate; preview capture failure should not imply the active Moonlight/Nova stream failed.

### DIFF
--- a/cmake/packaging/linux.cmake
+++ b/cmake/packaging/linux.cmake
@@ -39,6 +39,7 @@ set(CPACK_DEB_COMPONENT_INSTALL ON)
 set(CPACK_DEBIAN_PACKAGE_DEPENDS "\
             ${CPACK_DEB_PLATFORM_PACKAGE_DEPENDS} \
             debianutils, \
+            grim, \
             labwc, \
             libcap2, \
             libcurl4, \
@@ -59,6 +60,7 @@ set(CPACK_DEBIAN_PACKAGE_DEPENDS "\
             xwayland")
 set(CPACK_RPM_PACKAGE_REQUIRES "\
             ${CPACK_RPM_PLATFORM_PACKAGE_REQUIRES} \
+            grim, \
             labwc, \
             libcap >= 2.22, \
             libcurl >= 7.0, \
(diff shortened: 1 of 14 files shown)


papi-ux:
Smoke-tested this on the Linux headless labwc path using the PR build (`1.0.10.dbb41892`) with a Pixel 10 Pro client.

Live stream stats confirmed the new diagnostics are reporting the expected slow path clearly:

```json
{
  "streaming": true,
  "capture_path": "shm_cpu_capture",
  "capture_cpu_copy": true,
  "capture_gpu_native": false,
  "capture_transport": "shm",
  "capture_residency": "cpu",
  "capture_format": "bgra8",
  "encode_target_residency": "cpu",
  "encode_target_format": "p010",
  "fps": 59.8,
  "bitrate_kbps": 25987,
  "client_name": "Pixel 10 Pro"
}
```

The relevant log line now lines up with the structured stats:

```text
wlr: capture_transport=shm frame_residency=cpu frame_format=bgra8; capture will incur an extra CPU-side copy/conversion path
```

Also verified that the dashboard/web UI issue I hit during smoke testing was from the ad hoc binary-only install missing `/usr/local/assets/web/index.html`, not from this PR. Copying the built web assets restored the frontend without changing stream behavior.

One follow-up worth tracking separately: when Polaris is started from SSH/nohup, environment validation still warns that `WAYLAND_DISPLAY` is missing even though headless cage mode can create its own labwc socket. That warning is noisy in this smoke setup, but not a blocker for this diagnostics PR.

PR #46, opened 2026-05-04
Enable GPU-native cage probe for NVENC

papi-ux:
## Summary

- make the temporary windowed cage probe strict for the configured encoder and keep it from writing fallback probe results into the encoder cache
- allow the GPU-native cage probe to run when `linux_prefer_gpu_native_capture` is enabled, even before the deferred cage encoder has been selected
- resolve OpenGL shader assets from the executable-local `assets/shaders/opengl` directory before falling back to the installed assets path, and log missing shader sources explicitly

## Why

Headless labwc still lands on SHM/CPU capture. The runtime can preserve the GPU-native path by probing a temporary windowed labwc instance and, when NVENC validates there, starting the real cage session windowed instead of headless.

The smoke build also exposed a separate runtime asset issue: the GPU-native probe was failing GL shader linking because local builds looked for shaders under `/usr/local/assets`, even though the generated assets live beside the build executable.

## Validation

- `git diff --check`
- Linux CUDA smoke build: `cmake --build build-cuda -j$(nproc)`
- Live Retroid stream with `linux_prefer_gpu_native_capture=enabled`, `linux_use_cage_compositor=enabled`, `headless_mode=enabled`
- Confirmed active stream logs:
  - `session_manager: Windowed GPU-native cage probe succeeded with encoder [nvenc]`
  - `capture_transport=dmabuf frame_residency=gpu frame_format=bgra8`
  - `target_device=cuda target_residency=gpu target_format=nv12`
  - NVENC reported one encode session at 60 FPS


papi-ux:
Follow-up pushed in `1d1b430`.

What this adds:
- surfaces capture path, CPU-copy status, GPU-native status, and encode target in the dashboard and troubleshooting snapshot
- adds packaged user-service priority limits for capture/encode/audio workers (`LimitRTPRIO=95`, `LimitNICE=-10`)
- adds Quest 3 / Quest 3S high-throughput VR defaults
- documents the thread-priority warning and how to verify the active unit

Validation from the Linux smoke checkout:
- CUDA app build passed: `cmake --build build-cuda -j$(nproc)`
- generated service files include the new priority limits and passed `systemd-analyze verify --user`
- targeted C++ tests passed: `test_polaris_config`, `test_polaris_stream`
- web lint passed: `npm run lint`
- production web build passed: `npm run build -- --emptyOutDir`

One caveat: `npm run test:e2e -- tests/e2e/dashboard.spec.js` started but skipped because the remote non-interactive shell did not have `POLARIS_USER` / `POLARIS_PASS` loaded. The spec itself was updated to assert the GPU-native dashboard label when e2e is run against an authenticated UI.

PR #58, opened 2026-05-09
fix: attempt a GPU-native ext-image-copy DMA-BUF path before falling back to SHM

GregoryIvo:
Motivation

Reduce CPU-side frame transfers in headless cage/labwc sessions (reported in issue https://github.com/papi-ux/polaris/issues/49) by attempting a GPU-native ext-image-copy DMA-BUF path before falling back to SHM so incoming FPS and CPU usage can improve when supported.

Description

When headless RAM capture fallback is detected, opportunistically enable try_headless_extcopy_dmabuf if gpu_native_capture_supported and emit a one-time informational log; the change is in src/platform/linux/wlgrab.cpp and preserves the existing SHM fallback behavior if extcopy is unavailable.

Testing

No automated test suite was executed for this patch; the change was validated via local inspection and repository operations (git diff -- src/platform/linux/wlgrab.cpp, git status --short, git add/git commit) and a PR description was created with make_pr, and full CI/build and runtime verification are recommended.

### DIFF
--- a/src/platform/linux/wlgrab.cpp
+++ b/src/platform/linux/wlgrab.cpp
@@ -741,6 +741,19 @@ namespace platf {
         using_headless_ram_capture = true;
         try_headless_extcopy_dmabuf =
           cage_display_router::should_attempt_headless_extcopy_dmabuf(runtime_state);
+
+        // Even when GPU-native override is not active, opportunistically probe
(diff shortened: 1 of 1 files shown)


GregoryIvo:
I still need to give this a test, but thought id throw my hand into trying to find a fix.

PR #59, opened 2026-05-09
Updated how check boxes look

CattoSalad:
## Summary

I thought check boxes looked a bit off so I improved them to look like what I'd expect them to look like.
This is how I see them on current release:
image

Updated to look like this:
image
image
The value from this change is improved user experience.

## Testing

- [x] I updated docs or release notes when user-facing behavior changed. (unsure if needed)

## Notes
I may have missed something but I couldn't get the e2e tests to pass when I run `npm run test:e2e` maybe I am missing something. As it was expecting a different port, I was unsure how to get a development version running so I built and ran it as per https://github.com/papi-ux/polaris/blob/master/docs/building.md#build-and-install. That way I was able to validate that it all looks good. But happy to further discuss and test as needed.
I am not sure a change like this requires any docs updates.

I believe that would give the user a better experience, it's consistent with the current experience and matches both themes.

## Examples and reasoning
I am by no means a user experience expert, I merely went with what I thought made sense to me and felt functional. Happy to discuss all my changes.

I aimed to match the colour schemes and toggles from the quick controls on the dashboard.

### Space Whale:

image

### example:

image

### OLED Galaxy:
image

### example:

image

### Scales correctly to mobile devices too:
image

### before:
image

### after: 
image
It does make pages slightly longer but I think it makes it easier for the user.

Edit: had the wrong screenshots for the themes.

### DIFF
--- a/src_assets/common/assets/web/Checkbox.vue
+++ b/src_assets/common/assets/web/Checkbox.vue
@@ -105,18 +105,18 @@ const defValue = parsedDefaultPropValue ? "_common.enabled_def_cbox" : "_common.
 
 <template>
   <div :class="[props.class, { 'checkbox-field--compact': props.compact }]" :data-setting-key="props.id" class="checkbox-field">
-    <label :for="props.id" class="checkbox-field-control">
-      <div class="checkbox-field-toggle">
+    <label :for="props.id" class="checkbox-field-control flex items-start gap-2">
+      <span class="checkbox-field-toggle">
         <input type="checkbox"
                class="checkbox-field-input"
                :id="props.id"
                v-model="model"
                :true-value="checkboxValues.truthy"
                :false-value="checkboxValues.falsy" />
-      </div>
-      <span class="checkbox-field-copy">
-        <span class="checkbox-field-title-row">
-          <span class="text-silver cursor-pointer checkbox-field-label">
+      </span>
+      <span class="checkbox-field-copy flex min-w-0 flex-col gap-1">
+        <span class="checkbox-field-title-row flex flex-wrap items-baseline gap-x-2 gap-y-1">
+          <span class="text-silver cursor-pointer checkbox-field-label leading-snug">
             {{ $t(labelField) }}
           </span>
           <span class="text-xs text-storm checkbox-field-default" v-if="showDefValue">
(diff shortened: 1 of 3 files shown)


papi-ux:
Hey there! Thanks for this, actually think this direction looks useful and looks much cleaner too. 

Quick clean up before we can merge it!

 A couple things to fix if you don't mind:

  - The new `<div class="checkbox-field-toggle">` is inside a `<label>`, which is invalid HTML. Please switch that wrapper to a `<span>` or keep the input directly inside the label.
  - Since this replaces the native checkbox appearance, please add a visible keyboard focus state, e.g. `.checkbox-field-input:focus-visible`.
  - The branch now conflicts with `master`, mostly because `Checkbox.vue` was updated recently. Please rebase/update the branch and preserve the newer`master` layout classes.

After that i’m solid with another pass, and thanks again for any contribution (really), its genuinely solicited

CattoSalad:
Hey! Sure no problem, I'll get that done.

Honestly love this project, it just solves such a big problem for me with the virutal display on linux. I am just a full stack web developer by trade so I am trying to contribute however I can :)

CattoSalad:
@papi-ux all updated, .
Merge conflict has been sorted, now using a span inside the label and added the respective visible keyboard focus. I think it matches the theme. Here's a quick demo:

[Screencast_20260510_003516.webm](https://github.com/user-attachments/assets/c4cfd0b9-99e5-4dc1-8fe5-f853aec0c483)


I did notice this one:
image
I wanted to ask for your input on this, because it was not previously a checkbox, so I assume that's intentional, but  now it stands out as the odd one. Feels like it should be like the other options. If you'd like me to update it to the new checkbox or you have something else in mind, please let me know.
image

papi-ux:
The work is much appreciated and honestly, I agree with your suggestions. I like the new checkbox treatment better too.

For that odd one out: if it is semantically the same kind of boolean setting (which i think it is) and can use the shared `Checkbox.vue` path without custom behavior, I’m good with including it in this PR so the page feels consistent.

If it needs special behavior or a different component shape, let’s leave it for a follow-up. I’d rather keep this PR as “make checkbox fields better everywhere” instead of mixing in a separate settings-control refactor. After that, I’ll run a local lint/test/build pass and we should be close!

You rock thanks again!

CattoSalad:
Hello, just updated this. It uses the localisation copy now too, so I think there is a small improvement there too.
before:
image

after:
image

I also made sure the config updates correctly and I can confirm I still saw `mouse_cursor_visible = enabled` on the default config file `~/.config/polaris/polaris.conf`.

papi-ux:
awesome, thanks again! I merged this as #59 after checking the final merged result locally against current `master`

papi-ux:
Thanks again. I merged this as #59 after checking the final merged result locally against current `master`.

Validation passed:
- `npm run lint`
- `npm test`
- `npm run build`

The checkbox focus/wrapper fixes look good, and moving the host cursor setting onto the shared checkbox component fits the scope well.

PR #66, opened 2026-05-14
Add AMD GPU telemetry support and vendor-aware dashboard metrics

invider42:
## Summary

The dashboard only exposed GPU telemetry for NVIDIA cards via nvidia-smi, leaving AMD users with no metrics at all. This PR extends support to AMD GPUs using the amdgpu kernel sysfs interface (no ROCm dependency, works on any system with the standard kernel driver), and refactors the NVIDIA path into a clean vendor-dispatched structure so Intel or other vendors can be added later with minimal friction.

The vendor field and dynamic encoder label ensure the UI reflects what the hardware actually supports — NVENC is NVIDIA-specific and should not appear on AMD or Intel cards where it is meaningless.

- Extract NVIDIA GPU telemetry into query_nvidia_gpu()
- Add query_amd_gpu() using amdgpu sysfs (no ROCm required)
- Add detect_gpu_vendor() using active encoder type with filesystem fallback
- Expose gpu vendor field in system stats API response
- Add VCN encoder utilization for AMD RDNA2+ via vcn_busy_percent
- Resolve AMD GPU name via lspci -vmm with rocm-smi/amd-smi fallback
- Hide encoder gauge on GPUs that don't expose utilization
- Show NVENC or VCN label based on detected vendor

## Testing

- [x] I updated docs or release notes when user-facing behavior changed.

## Notes

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ AGENTS.md
 
 # build directories
 build/
+build-debug/
 build-local/
 build-test/
 build-tests/
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -4114,16 +4114,233 @@ namespace confighttp {
     send_response(response, tree);
   }
 
+
+#ifdef __linux__
+  /**
+   * @brief Detect the active GPU vendor.
+   *
+   * Prefers @c video::active_encoder_mem_type() as the primary signal since it reflects
+   * what the system actually uses for encoding. Falls back to driver/sysfs presence checks
+   * when no encoder has been chosen yet (i.e. before the first stream is started).
+   * @return "nvidia", "amd", or "" if the vendor cannot be determined.
+   */
+  static std::string detect_gpu_vendor() {
+    // amdgpu sysfs marker — used both for AMD detection and to distinguish AMD from Intel within vaapi
+    auto has_amdgpu = []() {
+      try {
+        for (auto &e : fs::directory_iterator("/sys/class/drm")) {
+          auto n = e.path().filename().string();
+          if (n.size() < 5 || n.substr(0, 4) != "card" || n.find('-') != std::string::npos) continue;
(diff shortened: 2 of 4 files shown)


papi-ux (changes requested):
Thanks for adding AMD telemetry support, i'm definitely limited since i only have NVIDIA, so this helps a bunch. The direction makes sense, and using amdgpu sysfs avoids requiring ROCm, which is the right baseline for most users.

I want to hold this before merging because the new AMD path can return a partial `gpu` object when optional sysfs values are unavailable. In that case the dashboard still assumes fields like `temperature_c`, `vram_used_mb`, and `vram_total_mb` exist, so AMD systems without every metric can render values like `undefined°` or `NaN GB`.

Could you update the dashboard to handle missing optional GPU metrics cleanly, and add test coverage for that case? In particular:

- Treat `temperature_c`, `power_draw_w`, clocks, VRAM, and encoder utilization as optional in the UI.
- Hide or render `--` for unavailable values instead of displaying `undefined` or `NaN`.
- Add/update a dashboard test with an AMD-style `gpu` payload that includes `vendor: "amd"` but omits at least `encoder_pct` and one of the VRAM/temperature fields.

Once that is covered, this should be in much better shape to merge.

invider42:
No problem, I've taken your feedback into account and made some changes, including adding a unit test.

papi-ux:
Thanks a bunch, this looks much better. The null-safe handling and hiding the encoder UI covers the main concern I had, and the new AMD dashboard test is exactly the kind of regression coverage I was looking for!

I’m gonna wait for/run the checks and take one more pass over the backend parsing before clearing the requested-changes review, but this addresses the feedback I left so you rock!

invider42:
Happy to help! Thank you as well

papi-ux (approved):

PR #149, opened 2026-06-28
Add openSUSE Tumbleweed build support (docs + optional CI)

Beary-Handsome:
## What

Adds first-class **openSUSE Tumbleweed** build support:

- **`docs/openSUSE.md`** — a from-source build guide (dependencies, build flags, optional RPM packaging).
- **`.github/workflows/opensuse-build.yml`** — a manual-trigger (`workflow_dispatch`) CI job that builds an installable `.rpm` + a relocatable tarball inside an `opensuse/tumbleweed` container and uploads them as an artifact.

## Why

There's no prebuilt openSUSE package today. This documents a clean, reproducible from-source path and lets anyone produce an openSUSE build/artifact.

## openSUSE-specific gotchas captured here

- **`-DBOOST_USE_STATIC=OFF` is required.** openSUSE ships only *shared* Boost. With the default static lookup, `find_package(Boost …)` reports missing component targets and the FetchContent fallback then collides with the partially-detected system targets (`add_library cannot create ALIAS target "Boost::headers" … already exists`). Building against shared system Boost avoids it.
- **RPM `Requires`.** The project's `CPACK_RPM_PACKAGE_REQUIRES` uses Fedora package names (`boost-filesystem`, `libcurl`, `xorg-x11-server-Xwayland`, …) that don't exist on openSUSE. The workflow overrides them with openSUSE names at `cpack` time; the shared-library dependencies are still auto-detected (autoreq), so `zypper` resolves Boost/libva/libcurl/etc. on its own.
- **Container builds** need `gawk` installed first — the `opensuse/tumbleweed` base image ships `busybox-gawk`, which blocks the real `gawk` that `desktop-file-utils`/`rpm-build` require.

## Notes

- Tested on Tumbleweed with an AMD GPU (VAAPI encode, `-DPOLARIS_ENABLE_CUDA=OFF`). NVIDIA hosts: add the CUDA toolkit and `-DPOLARIS_ENABLE_CUDA=ON`.
- The workflow is **manual-trigger only** (`workflow_dispatch`) so it doesn't impose container builds on every push — maintainers can add `push:` / `pull_request:` triggers if desired.

### DIFF
--- a/.github/workflows/opensuse-build.yml
+++ b/.github/workflows/opensuse-build.yml
@@ -55,6 +55,9 @@ jobs:
           #   * BOOST_USE_STATIC=OFF        - matches Arch; required on openSUSE (shared-only Boost)
           #   * CUDA=OFF, FAIL_ON_MISSING=OFF - matches Arch on a host without the CUDA toolkit
           #                                    (NVENC is NVIDIA-only; harmless on AMD/Intel)
+          #   * POLARIS_DESKTOP_ICON=polaris - upstream bug: the .desktop template references
+          #                                    @POLARIS_DESKTOP_ICON@ but never set()s it, so the
+          #                                    Icon= line ships empty and no launcher icon shows.
           cmake -B build -G Ninja \
             -DBUILD_DOCS=OFF \
             -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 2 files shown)


papi-ux (approved):
Approved after local verification.

PR #223, opened 2026-07-18
Per-app Display Source: isolated sessions (Family Mode), native EVDI capture, output pinning

Beary-Handsome:
## Summary

Adds a unified per-app **Display Source** to Polaris on Linux — one selector that answers "where do this app's pixels come from":

```
Display Source: default (primary display) | isolated session | virtual screen | specific output
```

All four sources follow the same lifecycle: acquire the display at session start (start the hidden compositor / create a virtual screen / power on the connector), capture it, release it at teardown.

Every feature here was field-tested end-to-end on real hardware (openSUSE Tumbleweed, KDE Plasma 6 Wayland, AMD RX 9070 XT + Ryzen iGPU, Moonlight client) — several commits are fixes for bugs that only surfaced on hardware.

## What's included (11 commits, reviewable independently)

**Per-app Display Source** (`960a3f1`)
- `display-source` per-app field with silent migration from legacy flags (`virtual-display`, plus the new `isolated-session` / `output-name`); legacy flags are normalized from the canonical value so downstream reads and external tools stay correct.
- **Isolated session (Family Mode)**: the app runs in the hidden headless compositor and streams alongside the live desktop — someone plays remotely while the host keeps using the PC. Implemented as a session-scoped save/restore override of the `linux_display` globals, so the existing session snapshot, display policy, cage lifecycle, gamepad isolation, and audio capture all behave correctly with no changes at those call sites. Audio is forced stream-only at the source and `/resume` preserves that against client `localAudioPlayMode` (prevents desktop audio leaking into the stream).
- **Specific output**: pins capture to a kernel connector name (HDMI dummy plug or EVDI screen). With `linux_auto_manage_displays`, the pinned output powers on at stream start and off at teardown — a dongle stays dark when not in use.

**Native EVDI frame capture** (`757bf85`, `06465da`, plus fixes)
- EVDI delivers frames to the process that opened the device (the DisplayLink consumer model); reading the EVDI card's scanout via KMS returns black even while the compositor renders the output correctly (verified against KWin compositor screenshots). New `evdigrab` backend pumps libevdi events (`request_update`/`grab_pixels`) into the CPU/BGRA encode path; cursor comes blended for free.
- Companion fixes found on hardware: openSUSE ships `libevdi.so.1` under `/usr/lib64/displaylink/` (off the linker path) — the dlopen candidate list now covers it (`fe9a59a`); `isLinuxVDisplayAvailable()` latched its first answer per-process, so a Polaris started before `modprobe evdi` reported "no backend" forever (`0ed44e8`); unprivileged processes can't create EVDI devices, so the failure now tells users about `modprobe evdi initial_device_count=1` (`c687e8a`); the generated EDID carried Apollo/Varjo branding ("Varjo Technologies Apollo Virtua" in KDE settings) — now `PLR` / "Polaris VDD", with spec-correct name-descriptor padding (`19bd70c`).

**Capture by connector name** (`701aa09`)
- kmsgrab targeted a numeric plane-enumeration index, which shifts when outputs hot-plug — exactly what virtual-display creation does. `process.cpp` already sets `output_name` to the created connector's name; kmsgrab now resolves non-numeric names to the connector's CRTC (kernel-style names via `drmModeGetConnectorTypeName`, matching `/sys/class/drm` and kscreen). Numeric behavior unchanged; connected connector names are appended to `display_names()` after all numeric entries so position invariants hold.

**Headless quality** (`c2c8dd7`, `96473b9`)
- Headless labwc used GLES2, which leaves GPU-accelerated clients (Steam CEF, Proton/native games) black off-screen. Now Vulkan on AMD/Intel (auto-detected from the DRM driver under `/sys/class/drm`); NVIDIA keeps GLES2 (Vulkan headless SEGVs on nvidia + wlroots 0.19).
- Optional `swaybg` splash backdrop during game cold-start (user image at `~/.config/polaris/splash.png`, solid color fallback, no hard dependency).

**Virtual Screen Backend** (`86cf63d`)
- `headless_source = auto | virtual | evdi` overriding backend auto-detection, cached in the existing detection TTL keyed by source. A reserved `physical` value parses for compatibility and falls through to auto (real-connector capture is the output pin's job).

## Review question for the maintainer

While porting I noticed current master's `platf::display()` (linux `misc.cpp`) selects capture purely from the startup source bitset — there's no cage-first dispatch. On our older base, a per-app cage session whose startup bitset chose KMS captured the *physical desktop* instead of the cage (privacy leak); we fixed it with a cage-first branch. Your session-generation rework may have solved this differently — flagging rather than re-adding it blind. Happy to add the guard here if it's still reachable.

## Validation

- Built green against current master via the openSUSE Tumbleweed workflow.
- Hardware-tested on the fork (same code): Family Mode with Steam Big Picture + native-Linux and Proton titles, concurrent desktop use, audio/input isolation; EVDI create→render→native-capture→stream; dongle pin with auto power on/off.
- Known limitation (pre-existing, not introduced here): absolute mouse input maps to the primary output on KDE Wayland for extended-screen sources (uinput abs devices are compositor-mapped). Tracked as follow-up; the right fix is likely libei.

Co-authored with @Beary-Handsome, who did all the hardware validation.


Beary-Handsome:
Closing to revise — splitting the isolated-session (Family Mode) work from the EVDI/dongle display-source work after more hardware testing. Will reopen a focused PR.

papi-ux:
amazing, great call. appreciate the efforts!

Beary-Handsome:
> amazing, great call. appreciate the efforts!

Family mode works great, and I'm happy with how it came out.

I tried extensively to break it, I couldn't. True separation of main device and streaming - works flawlessly.

EVDI on the other hand, I was able to break. It does work really well, but I would like to do further extensive testing and mitigate any fragility before it appears in another PR.

PR #224, opened 2026-07-18
Per-app isolated sessions (Family Mode) for Linux

Beary-Handsome:
## Summary

Adds a per-app **Isolated Session (Family Mode)** to Polaris on Linux: mark an app and it launches inside Polaris's hidden headless labwc compositor, concurrently with the live desktop. A second person streams a game to another device while the primary user keeps using the PC — the streamed player controls only the game, the desktop user's physical keyboard/mouse keep controlling the desktop, and game audio is routed to a private virtual sink so it never plays on the host speakers.

This is the focused follow-up to #223, which I closed to split the solid isolated-session work out from more experimental EVDI/display-source changes that need more hardware testing.

## Commits (reviewable independently)

- **feat(linux): per-app isolated sessions (Family Mode)** — the feature.
- **fix(family-mode): Vulkan renderer for headless labwc on AMD/Intel** — headless GLES2 leaves GPU-accelerated clients (Steam's CEF, Proton/native games) black off-screen; select Vulkan on AMD/Intel (auto-detected from the DRM driver under /sys/class/drm), keep GLES2 on NVIDIA where headless Vulkan is unstable on wlroots 0.19.
- **feat(family-mode): swaybg splash backdrop** — optional dark backdrop during the game's cold-start instead of a black void (no hard dependency; solid-color fallback).

## Design

Implemented as a **session-scoped override of the linux_display globals** at the top of `execute_impl()`: when the launched app opted in, `headless_mode` and `use_cage_compositor` are forced true and `prefer_gpu_native_capture` false (keeping the compositor genuinely off-screen). Every downstream read — the session snapshot, display policy, cage lifecycle, gamepad isolation, encoder-probe deferral, audio capture — then behaves as a headless+cage session with **no edits at those ~dozen call sites**.

Saved values are restored from `terminate_impl()` on both the success and `fail_guard` paths, plus a scope guard covering the window before the main `fail_guard` is armed — and only when the app opted in, so normal sessions are never touched. Audio isolation is forced at the source (`launch_session->host_audio` + cleared explicit sink), and `/resume` preserves it against the client's `localAudioPlayMode` so a reconnect can't leak desktop audio.

## Testing

Field-tested end-to-end on openSUSE Tumbleweed, KDE Plasma 6 Wayland, AMD RX 9070 XT, Moonlight client: Family Mode with Steam Big Picture launching native-Linux and Proton titles, concurrent desktop use (desktop stays usable, no focus/edit-mode disruption), audio split verified via `pactl` (game on the virtual sink, desktop audio on the real default), input isolation (streamed input to the game only; host kbd/mouse to the desktop), and single-session refusal. Built green against current master via the openSUSE Tumbleweed workflow.

Co-authored with @Beary-Handsome, who did all the hardware validation.

### DIFF
--- a/src/confighttp_validation.cpp
+++ b/src/confighttp_validation.cpp
@@ -384,6 +384,7 @@ namespace confighttp::validation {
       "elevated"sv,
       "exclude-global-prep-cmd"sv,
       "exclude-global-state-cmd"sv,
+      "isolated-session"sv,
       "launching"sv,
       "per-client-app-identity"sv,
       "terminate-on-pause"sv,
--- a/src/nvhttp.cpp
+++ b/src/nvhttp.cpp
@@ -5278,6 +5278,14 @@ namespace nvhttp {
     if (no_active_sessions && args.find("localAudioPlayMode"s) != std::end(args)) {
       host_audio = util::from_view(get_arg(args, "localAudioPlayMode"));
     }
+#ifdef __linux__
+    // A resumed isolated session (family mode) must keep stream-only audio: the
+    // client's localAudioPlayMode would otherwise re-enable host audio and leak
+    // the desktop's sound into the stream.
+    if (proc::proc.isolated_session_active()) {
+      host_audio = false;
+    }
+#endif
     auto launch_session = make_launch_session(host_audio, false, args, named_cert_p.get());
     const bool watch_only = launch_session->watch_only;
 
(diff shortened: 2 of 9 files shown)


Beary-Handsome:
Superseded by #226, which consolidates the full Headless Streaming Display feature (EVDI native capture + Family Mode + capture-follow + dongle swap + EVDI-as-primary) into one PR. The changes here are included there. Closing to keep review in one place.

PR #226, opened 2026-07-18
Headless Streaming Display for Linux/KDE (EVDI native + dongle swap + EVDI-as-primary; supersedes #224, #225)

Beary-Handsome:
## Headless Streaming Display for Linux/KDE — script-free display management

This consolidates all the Linux headless-streaming work into one coherent feature: stream
a game host **without a dedicated physical display**, and have Polaris manage the display
swap for you instead of hand-rolled `kscreen-doctor` prep-cmd scripts. It supersedes and
bundles #224 (Family Mode) and #225 (capture-follow); those are closed in favor of this PR.

Everything here is **opt-in** — default behavior is unchanged, and no new required config
keys are introduced for the swap (it reuses the existing `headless_source` /
`headless_swap_primary` / `linux_auto_manage_displays` / `linux_primary_output` settings).

### What's in it (bottom to top)

**EVDI native capture** — consume EVDI virtual-display frames directly via `libevdi`
(KMS reads of an EVDI plane come back black), with an avcodec encode device from the
grab backend. Finds `libevdi.so.1` including openSUSE's off-path
`/usr/lib64/displaylink/` location, gives actionable guidance when no EVDI device can be
opened, and stops latching virtual-display availability per-process (it changes at
runtime as the module is loaded/unloaded).

**KMS capture by kernel connector name** — target a capture output deterministically by
its connector name (e.g. `HDMI-A-1`), not just a plane index.

**Family Mode (per-app isolated sessions)** — launch a game in an isolated nested `cage`
compositor so it captures the game, not the physical desktop; Vulkan renderer for headless
labwc on AMD/Intel; a splash backdrop over the cold-start black gap; and a fix so capture
targets the cage, not the physical desktop.

**`headless_source` selector + per-app Display Source** — choose the virtual-screen
backend (`auto` / `virtual` / `evdi` / `physical`) globally and per-app.

**capture-follow (crash + freeze fix)** — when a display-topology change retires the
pinned plane (monitor off, desktop moved to another CRTC), re-resolve onto the now-active
output instead of dereferencing a null plane (crash) or spinning on `timeout` forever
(frozen stream). ~12 lines in `kmsgrab.cpp`, helps every KDE user regardless of the rest
of this PR.

**Headless Streaming Display — the swap:**
- *Physical dongle* (`headless_swap_primary`): at stream start, move the desktop onto a
  real dummy-plug connector and disable the physical monitor; restore on teardown. The
  built-in replacement for manual swap scripts.
- *EVDI-as-primary* (the same toggle with `headless_source=evdi`): create the EVDI even
  under auto-manage, confirm it is a live, addressable kscreen output, promote the desktop
  onto it, capture via the CRTC-independent evdigrab path, and un-promote (restore the
  physical monitor) **before** destroying the EVDI. **Headless streaming with no hardware
  at all.**

### Safety / design notes
- One canonical **kernel connector name** drives both the kscreen swap and capture routing
  (KScreen names outputs by kernel connector, so `find_evdi_output()`'s value is the swap
  target).
- Every EVDI-promotion failure path (create failed, non-EVDI backend, unresolved
  `VIRTUAL-1` name, EVDI never appears in kscreen, unset `primary_output`) **tears the EVDI
  down and falls back to capturing the real desktop** — never a stranded/black desktop,
  never a disabled monitor left un-restored. Promotion also defers to a per-app output pin.
- Teardown un-promotes atomically (one `kscreen-doctor` call, so KScreen never sees zero
  enabled outputs) with a settle before EVDI destroy.
- **Family Mode is exempt** from the swap (cage-gate), so isolated sessions never touch the
  host's monitor — verified even with `headless_source=evdi` set globally.
- `headless_swap_primary` guards against `streaming_output == primary_output`.

### Validation
Hardware-tested on **openSUSE Tumbleweed, AMD RX 9070 XT, KDE Plasma 6 (Wayland)**:
- ✅ EVDI-as-primary: desktop promoted onto the virtual `DVI-I-1`, physical monitor
  disabled, streamed via native evdigrab, restored on teardown (log-confirmed
  `promoting desktop onto [DVI-I-1]` + atomic un-promote).
- ✅ Physical-dongle swap: desktop moved onto the dongle and restored cleanly.
- ✅ Family Mode: host monitor never dimmed, even with `headless_source=evdi` globally.
- ✅ EVDI native capture and the capture-follow reinit path validated in earlier testing.

Reviewed with layered adversarial passes (design + implementation) before each hardware
test; all abort/edge paths were hardened to fall back to the physical desktop rather than
stream black.


luxus:
i tried evdi myself.. so far i know its not possible to enable hdr on it right?

Beary-Handsome:
> i tried evdi myself.. so far i know its not possible to enable hdr on it right?

As far as I'm aware you're right.

papi-ux:
Hey! First off i'm really sorry this one sat for so long, that's completely on me. Good news though! Your design did make it in, just not as this PR. When the Linux stream paths got reworked at the end of July, the dongle swap went in as its own first-class path instead of a separate mode. There's a SWAP_PRIMARY in stream_path.h literally commented "#226-style" pointing right back at you, the headless_dongle path does the desktop swap + privacy blanking + restore on stop, and EVDI lives in virtual_display.cpp (dlopen'd so it stays a soft dependency). Which means rebasing 1500+ lines across 21 files onto an architecture that didn't exist when you wrote this would land you right back at what's already shipping sooo i really don't want to put you through that, so i'm gonna close it out here. 

But genuinely, the privacy blanking and the restore-on-stop behavior are yours, and the #224 to #225 to #226 trail shows how much you iterated to get there. If any of it doesn't work the way you meant it to on your hardware (EVDI-as-primary especially, that's the piece i'm least confident carried over cleanly), open an issue and i'll treat it as a regression against your design, not a new request. Thanks so much for the work here, and sorry again for the quiet! Cheers

PR #229, opened 2026-07-24
linux: wait for Steam singleton release before private stream launch

atgehrhardt:
## Problem

When starting a private stream while the desktop Steam client is running, Polaris closes desktop Steam first (`request_desktop_steam_shutdown_for_private_stream()`). That function reported success as soon as no `steam`/`steamwebhelper` processes remained in `/proc`.

Process exit is not the same as teardown completing. Steam's client singleton is the `~/.steam/steam.pipe` FIFO: every `steam ...` invocation — including the `steam steam://rungameid/...` used to launch the game in the private session — forwards its arguments to whatever instance is reading from that pipe and then exits. When the private launch raced ahead while the dying instance still held the pipe, the launch URI was forwarded to the dying instance and swallowed: the stream started, but the game never launched. Quitting Steam manually (and waiting) avoided the race, which is how this was reproduced/diagnosed.

## Fix

After the process scan reports Steam gone, additionally wait (up to 5 s, same polling cadence) for the `steam.pipe` FIFO to have no listener before reporting the shutdown as complete. Opening the FIFO `O_WRONLY|O_NONBLOCK` fails with `ENXIO` once no instance is reading — exactly the point where a new instance will own the singleton and process the launch URI itself. `ENOENT` (no pipe at all) is also treated as released; any other error fails closed, matching the existing style of the surrounding code.

Verified empirically against a live Steam client: while Steam runs, the pipe has a listener; the flock on `steam.pid` hypothesis was tested and ruled out (Steam does not lock that file).

## Testing

- New unit test `SteamInstancePipeDetectsSingletonListener` in `test_process_migration.cpp` covers: FIFO with no reader (released), FIFO with a reader in a forked child (held), reader killed (released), missing path (released).
- `test_polaris_config`: 236/237 pass. The single failure (`ProductionPreCageSteamTeardownUsesImmutableExactGeneration`) also fails on unmodified `master` on the same host — it is caused by a live desktop Steam client running during the test, not by this change.
- Manually verified end-to-end on Fedora 44 (CUDA + browser stream enabled): starting a private stream with desktop Steam running now launches the game reliably.

### DIFF
--- a/src/process.cpp
+++ b/src/process.cpp
@@ -1709,6 +1709,34 @@ namespace proc {
     thread_local pid_t forced_desktop_steam_scan_only_pid = -1;
 #endif
 
+    // Steam's client singleton is a FIFO at ~/.steam/steam.pipe: every
+    // `steam ...` invocation (including `steam steam://rungameid/...`)
+    // forwards its arguments to whatever instance is reading from that
+    // pipe and then exits. The last Steam process can vanish from /proc
+    // while teardown still holds the pipe, so process scanning alone
+    // cannot tell when a new instance may safely own the singleton; a
+    // launch URI forwarded in that window is swallowed by the dying
+    // instance and the game never starts.
+    bool steam_instance_pipe_listener_active(const std::string &pipe_path) {
+      const int fd = open(pipe_path.c_str(), O_WRONLY | O_NONBLOCK | O_CLOEXEC);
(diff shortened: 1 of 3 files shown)


atgehrhardt:
Closing in favor of combining this with another PR I will be submitting soon. Didn't make sense to separate.

PR #287, opened 2026-08-06
feat(library): let a curated artwork identity decide the estimate

papi-ux:
## Summary

A wrong completion estimate was visible but not fixable. #282 carries `matched_name` so a
client can show which game an estimate is actually for, but seeing it was the whole of it
— correcting it meant hand-editing `beat_times.json` on the host.

The concept answers its own question here:

> the artwork studio already resolves a game to a canonical identity through
> `PolarisArtworkMatchCandidate`. If you have told Polaris that this entry **is** Slay the
> Spire 2 for artwork purposes, that is the same answer HLTB needs — the match should be
> shared rather than resolved twice, and corrected in one place.

So a **manual** artwork match now decides which game the estimate is for. Two calls in that
worth stating:

- **It outranks the Steam app id.** The id is what made a wrong estimate confident in the
  first place, so a correction the id overrules is not a correction.
- **Only a manual match counts.** An automatic one is the same guess the estimate would
  make on its own, and treating it as authority would launder a guess into a fact.

The id still travels with the lookup, so a corrected answer replaces the wrong entry
rather than sitting beside it.

## Testing

`tests/test_polaris` — 184 tests, 0 failures.

Verified on a Retroid against this host: Slay the Spire 2 reads its estimate for `Slay the
Spire II`, and the client line that admits the mismatch now lands on Artwork Studio for
that entry, where the identity is set. Correcting it there is what changes the estimate.

## Notes

No behaviour change for any entry whose artwork match is automatic, which is all of them
until somebody curates one. No changes to pairing, certificates, trusted subnets, client
commands, shell hooks, packaging, or privilege boundaries.

### DIFF
--- a/src/beat_times.cpp
+++ b/src/beat_times.cpp
@@ -293,6 +293,16 @@ namespace beat_times {
     return by_name->second;
   }
 
+  std::optional<estimate_t> lookup_for_identity(const dataset_t &data,
+                                                std::string_view curated_title,
+                                                std::string_view steam_appid,
+                                                std::string_view name) {
+    if (!curated_title.empty()) {
+      return lookup(data, "", curated_title);
+    }
+    return lookup(data, steam_appid, name);
+  }
+
   std::string match_key(std::string_view name) {
     std::string folded;
     folded.reserve(name.size());
--- a/src/beat_times.h
+++ b/src/beat_times.h
@@ -101,4 +101,21 @@ namespace beat_times {
    */
   std::optional<estimate_t> lookup(const dataset_t &data, std::string_view steam_appid, std::string_view name);
 
+  /**
+   * @brief The estimate for a game somebody has told us the identity of.
+   *
+   * A curated title is a correction, and the app id is usually what made the wrong
+   * estimate confident in the first place, so a curated entry never falls back to it:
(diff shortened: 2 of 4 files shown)


papi-ux:
Rebased onto `a20dc5d` and pushed a fix commit — the curated identity fell back to the app id, which undid the correction in the case the feature exists for.

`lookup(data, "", curated)` only searches `by_name`. When the curated title is one the dataset does not carry yet — which is the case somebody is most likely correcting — it missed, the fallback found the estimate filed under the app id, and the client was shown an estimate for the game the correction had just rejected. Because that fallback returned a value, `request_lookup` was never reached either, so the curated title was never asked about in the background and the correction could never take effect at all.

That contradicted this PR's own stated rule: *a correction the id overrules is not a correction.*

A curated entry now answers with its own title or answers with nothing. The rule moved into `beat_times` as `lookup_for_identity`, next to the id-over-name rule it overrides, so it can be tested against a dataset instead of through the HTTP handler.

Three tests added. A mutation probe restoring the original fallback fails exactly `ACuratedTitleTheDatasetLacksAnswersWithNothing` and nothing else.

Verified on pc-papi: full `ninja` exit 0, `ctest` 17/17, `BeatTimesTests` 15/15.

PR #294, opened 2026-08-07
chore(nix): align gamescope-polaris with +polhdr patch stack

luxus:
## Summary
- Replace fragmented gamescope HDR companion patches with a focused three-patch surface: 10-bit BT.2020 PQ formats, optional cursor composite, and version stamp (`+polhdrN`).
- Wire the new patches through `gamescope-polaris` packaging and archive superseded local patches.
- Document the patch set in `nix/patches/gamescope/README.md`.

## Stack
**PR2 of 3** (merge first)

Suggested order: **PR2 → PR1 → PR3**

| PR | Branch | Role |
|----|--------|------|
| **this** | `feat/gamescope-polaris-pf-align` | gamescope patches |
| next | `feat/linux-stream-sink-claim` | stream sink claim |
| last | `feat/gamescope-session-lifecycle` | session + HDR capture |

## Test plan
- [ ] `gamescope-polaris` builds on NixOS (lea)
- [ ] Idle gamescope starts; version stamp present if expected
- [ ] No regression on non-HDR stream attach


papi-ux:
Hey again! Closing this one, but same story as the other two, work landed, just not as this PR.                                                                               

It's all in #305, rebased straight off master. One thing i had to fix on the way through, and it's worth knowing about:                                                      

Patch 10 couldn't apply. Its last hunk header said @@ -2335,17 +2335,32 @@ but the body actually had 33 new-side lines, not 32, so patch bailed with "malformed patch at line 260" and the whole stack died at the very first patch. The diff content was completely fine, it was purely the header arithmetic. Regenerated it from the same content with your authorship preserved.                                                                                                                                              

Here's the part that was on me, not you: nothing in our CI built nix. Not one workflow even mentioned it. So this PR sat there showing nine green checks while every single file it changed was untested.                                                                                                                                                            

So we fixed that too. #308 just landed: the patch stacks now get checked on every push, and CI actually builds gamescope-polaris. Your install check ran for the first time today!                                                                                                                                                                

```
gamescope-polaris> Running phase: installCheckPhase                                                                                                                          
[gamescope] gamescope version 0-unstable-2026-08-01+polhdr2 (gcc 15.2.0)          
```                                                                                           

That whole build is 5 minutes. It should have been there since the beginning. Really nice work! consolidating 01/04/07 into one patch and adding the version stamp so the build can prove itself is a much better shape than what we had. Your PR is what made us get here so thanks again! Wow, look at us

PR #295, opened 2026-08-07
fix(linux/audio): claim stream sink as session default

luxus:
## Summary
- Claim the Polaris stream sink as the WirePlumber/Pulse session default while streaming.
- Stop preferring EasyEffects as the capture target over the virtual stream sink.
- Drop session re-pin thrash; let linking follow the configured default.
- Soften gamescope session shell audio pin loop accordingly.
- Base `release_default_sink()` returns not-implemented so platforms without claim still restore the host sink after the stream.

## Stack
**PR1 of 3** — merge **after** #294

| Order | PR | Branch |
|------|-----|--------|
| 1 | #294 | `feat/gamescope-polaris-pf-align` (patches) |
| 2 | **this** | `feat/linux-stream-sink-claim` (audio) |
| 3 | session | `feat/gamescope-session-lifecycle` |

This branch is stacked on #294, so the PR diff currently includes the gamescope patch commit until #294 lands. Review focus: **audio claim** commit only.

## Test plan
- [ ] Start stream with EasyEffects as host default → capture uses stream sink; apps route to claim
- [ ] End stream → previous default restored
- [ ] Unit tests: sink selection / claim policy


papi-ux:
Hey! Closing this one out, but the work absolutely landed, just a wee bit unstacked.                                                                                                  

The problem was the stack itself though i found... everything here sat behind #294, so none of it could move while i was still working out what to do about the vendored patches. But the audio claim and the session lifecycle don't actually depend on those patches at all, although they cherry-picked straight onto master with zero conflicts. So i pulled them out and landed them separately, with your authorship intact of course:                                                                                                                                     

- #301  the stream sink claim                                                                                                                                               
- #302 gamescope session lifecycle + the HDR format negotiation                                                                                                            
- #305 the patch stack itself, rebased off the others                                                                                                                      

Two things i fixed on the way through, both worth knowing about:                                                                                                             

1. On the audio claim, restore_sink was doing double duty. So a session that never claimed still called `release_default_sink()` and decremented a claim another session was holding. With two concurrent sessions that can restore the host default underneath a stream that's still running.                                                                                                                                                               

2. Aaand patch 10's last hunk header said 32 new lines when the body had 33, so patch bailed with "malformed patch at line 260" and the whole stack died at the first patch.      

Content was solid, just the arithmetic needed a tweak. Regenerated it with your authorship preserved.                                                                                        

That one's really a lesson for me, not you: nothing in our CI builds nix, so all nine checks went green on a PR whose every file was untested. That's on our setup, not your patch.                                                                                                                                                                       

For real though, genuinely great work, and the HDR negotiation following the encode rather than format ordering is the right call. Cheers!

PR #296, opened 2026-08-07
fix(linux): harden gamescope session lifecycle and HDR capture

luxus:
## Summary
Make `gamescope_stream` teardown and reconnect reliable after nested WSI and attach launches, and align PipeWire format negotiation with whether the client stream actually encodes HDR.

### Session lifecycle
- Harden wait/ownership recovery and nested stop restore-idle handoff
- Recover when nested marker is already dead instead of looping forever
- Finish owner cancel when attach Steam bwrap is not a private session leader
- Force-clear incomplete gamescope attach generation so the next launch is not refused until polaris restarts

### HDR / portal capture
- Prefer exclusive `xBGR_210LE` PQ EnumFormat when force **and** client `dynamicRange` request HDR on `gamescope_stream`
- Prefer exclusive 8-bit BGRx when the stream is SDR so PQ capture cannot feed an SDR encode (bright-red wash)

Also refreshes `polaris-stream-ui` `npmDepsHash` and adds stack plan/handoff docs.

## Stack
**PR3 of 3** — merge **last**, after #294 and the audio claim PR

| Order | PR | Branch |
|------|-----|--------|
| 1 | #294 | patches |
| 2 | #295 | `feat/linux-stream-sink-claim` |
| 3 | **this** | `feat/gamescope-session-lifecycle` |

This branch is stacked on the audio claim branch (which includes #294). Diff includes parent stack commits until those PRs land; review focus: **session + portal HDR** commit.

## Field-tested on lea
- HDR on: `prefer_hdr`, spa `xBGR_210LE`, `dst_p010=true`
- HDR off: `prefer_sdr`, spa `BGRx`, `src_xb30=false`
- Cancel / next-game launch after incomplete attach cleanup no longer bricks launches

## Test plan
- [ ] Nested WSI (client HDR on) start/stop STS2 or BG3; idle restores
- [ ] Attach path (client HDR off) start/stop; colors not washed red
- [ ] Quit from Moonlight (owner cancel) then immediately launch another app
- [ ] After “close session when game ends”, next title starts without host restart
- [ ] Shell: `test_gamescope_session_stop_shell.sh`; portal policy source contract tests


papi-ux:
Hey! Closing this one out, but the work absolutely landed, just a wee bit unstacked.                                                                                                  

The problem was the stack itself though i found... everything here sat behind #294, so none of it could move while i was still working out what to do about the vendored patches. But the audio claim and the session lifecycle don't actually depend on those patches at all, although they cherry-picked straight onto master with zero conflicts. So i pulled them out and landed them separately, with your authorship intact of course:                                                                                                                                     

- #301  the stream sink claim                                                                                                                                               
- #302 gamescope session lifecycle + the HDR format negotiation                                                                                                            
- #305 the patch stack itself, rebased off the others                                                                                                                      

Two things i fixed on the way through, both worth knowing about:                                                                                                             

1. On the audio claim, restore_sink was doing double duty. So a session that never claimed still called `release_default_sink()` and decremented a claim another session was holding. With two concurrent sessions that can restore the host default underneath a stream that's still running.                                                                                                                                                               

2. Aaand patch 10's last hunk header said 32 new lines when the body had 33, so patch bailed with "malformed patch at line 260" and the whole stack died at the first patch.      

Content was solid, just the arithmetic needed a tweak. Regenerated it with your authorship preserved.                                                                                        

That one's really a lesson for me, not you: nothing in our CI builds nix, so all nine checks went green on a PR whose every file was untested. That's on our setup, not your patch.                                                                                                                                                                       

For real though, genuinely great work, and the HDR negotiation following the encode rather than format ordering is the right call. Cheers!

PR #308, opened 2026-08-07
ci(nix): check the vendored patch stacks and build what they patch

papi-ux:
## Summary

`nix/` is the one directory in this repo that no CI job looks at. #294 changed nine files under it, collected nine green checks, and not one of those checks read a line it changed — the stack it added could not apply at all, because patch 10's last hunk header claimed 32 new-side lines while its body had 33. `patchPhase` applies the declared list in order and stops at the first failure, so a one-line arithmetic slip took the whole gamescope stack with it.

This closes that gap in three tiers, scoped by what a change can actually break.

**`scripts/check-nix-patches.py`** — the properties that need neither network nor nix: every hunk header against its own body, every patch on disk either declared by a package or parked under `archive/`, every declared patch present. With `--apply` it also fetches each package's pinned upstream and applies that package's stack in the declared order using `patch -p1`, the same tool `patchPhase` uses and the one whose tolerance for line offsets the build depends on. Declared order rather than sorted: `gamescope-polaris` applies 10, 11, 12, 02, 03, 06, and a numeric sort would test a stack nobody ships.

**`.github/workflows/nix.yml`** — `nix flake check` plus a per-package eval, then `nix build` of `gamescope-polaris` and the portal, scoped by path to changes that can reach them. `polaris-stream` (CUDA, npm, the whole C++ tree) is manual plus weekly instead.

The gamescope build is the piece that matters most. The package carries a `doInstallCheck` that greps `--version` for the `+polhdr` stamp patch 12 adds, written precisely so a stack that applies but does not take fails the build rather than someone's session. It can only run inside a build, and nothing built the package, so it had never run once.

**Public Hygiene** runs the offline half on every push. It costs about a second, so there is no reason to gate it on touching `nix/`.

**Removed** `01`, `04` and `07` from `nix/patches/gamescope/`. They were left beside the live patches as byte-identical copies of the archived ones when 10 superseded them; a patch sitting among the live ones without being applied reads as live, which is what the declaration check calls an error. `archive/` keeps them.

## Exact candidate

`a5ab467`

## Verification

Against master, on a Linux host:

- `scripts/check-nix-patches.py` on master before this change reported the three unreferenced patches, and nothing else — the ten patch files on master are all internally consistent.
- Reintroducing #294's exact miscount (`+2335,33` → `+2335,32` in patch 10) produced:
  ```
  nix/patches/gamescope/10-pipewire-offer-10-bit-BT2020-PQ.patch:224:
    `@@ -2335,17 +2335,32 @@` claims 17 old / 32 new lines, body has 17 old / 33 new
  ```
  and `patch -p1` on the same file fails with `malformed patch at line 260`, which is what #294 hit.
- `--apply` fetches both pinned revisions and applies all seven vendored patches clean: gamescope at `ff6b924` (10, 11, 12, 02, 03, 06) and the portal at `412a4bf` (01). Whole run, both downloads included, 0.8s.
- `patch` exits 1 when it cannot find a target file, so the apply check cannot pass vacuously — confirmed against an empty tree.
- All four Public Hygiene scripts pass on the branch.
- Both workflow files parse; `nix.yml` yields the four expected jobs and four triggers.

Not verified locally: the `nix` steps themselves. There is no nix on MacPapi or pc-papi, which is part of how this gap survived. The eval and build jobs run here first.

## Notes

`nix/README.md` now states what is covered and what is not. Still uncovered: the home-manager module (`homeModules` is not an output `nix flake check` knows about) and evaluation for `aarch64-linux`, which is skipped deliberately — this flake turns CUDA on for every system, so `--all-systems` would report a red for a platform nobody here builds for.

Nix is installed with the upstream installer rather than a third-party action, since every other workflow here uses GitHub-owned actions only.


papi-ux:
The nix steps have now run, so replacing the "not verified locally" note above with what actually happened.

**The install check ran, and passed for the right reason.** From the build log:

```
gamescope-polaris> applying patch /nix/store/…-12-polaris-stamp-version-polhdrN.patch
gamescope-polaris> Running phase: installCheckPhase
[gamescope] gamescope version 0-unstable-2026-08-01+polhdr2 (gcc 15.2.0)
```

`+polhdr2` is the whole point: the stack applied *and* took. That assertion was written in #294 and has never executed until now.

**It is far cheaper than budgeted.**

| Job | Time |
|---|---|
| Patch stacks (no nix) | 7s |
| Flake eval | 28s — of which nix installs in 3s |
| Build gamescope-polaris | 5m12s |
| Public Hygiene (whole suite) | 10s |

The 90-minute timeout on the build stands as a safety net, but `cudaSupport = true` does not drag gamescope's closure into a mass rebuild the way I expected — nearly all of it comes from cache.nixos.org.

**`nix flake check` confirmed both coverage gaps** rather than me having to assume them:

```
warning: unknown flake output 'hjemModules'
warning: The check omitted these incompatible systems: aarch64-linux
```

Both are now written down in `nix/README.md`. Worth noting the warning names `hjemModules` only — nix recognises `homeModules`, so home-manager is less uncovered than the README's older note suggested.

**The `gamescope-hdr` alias is real, not nominal.** The per-package eval resolves it and `gamescope-polaris` to the same derivation, `85wlm4vh0lzd4g7h4cphv0b6pz97d8qn`.

PR #379, opened 2026-08-12
feat(api): make the Doctor's safer-bitrate action executable from the web

papi-ux:
## Summary

Closes the loop the Mission Control redesign left open: the Doctor's lower_bitrate safe action targeted the game-stream server's client-settings route, unreachable from a browser session, so the web rendered the host's own recommendation as inert "Suggested:" text.

- New `POST /api/clients/settings` on the web server (session auth + CSRF): a scoped patch for a paired client's stream tuning (`target_bitrate_kbps`, `display_mode`), same validation ranges as the game-stream handler.
- Backed by `nvhttp::patch_client_stream_settings`: clone-and-replace under the client-state lock, absent fields keep their values, so a partial caller can never reset the client's name/permissions/commands (the hazard a bare `/api/clients/update` call carries). Display-mode format validation exposed as `nvhttp::parse_display_mode_selection`.
- `doctor_safe_action` emits the new endpoint for lower_bitrate; the dashboard merges the active session's client uuid into the payload. The web already gates execution on `/api/` endpoints, so the inert text upgrades to a working confirm-gated button with no further UI change. Nova's `/polaris/v1/client-settings` channel is untouched.

## Exact candidate

`70a819a` on `doctor-web-action`. 6 files, +182/-2.

## Verification

- New unit assertion in `test_stream_stats.cpp`: an encoder_load verdict's action targets `/api/clients/settings` with a `target_bitrate_kbps` payload (POST).
- Web suite 302/302, eslint clean, budgets pass (dashboard change is the uuid merge only).
- C++ compile + stream-stats suite via the CI distro matrix and sanitizer job.

### DIFF
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -3269,6 +3269,11 @@ namespace confighttp {
       }
 
       const auto result = nvhttp::patch_client_stream_settings(uuid, target_bitrate_kbps, display_mode);
+      if (result == nvhttp::client_mutation_result_t::success && target_bitrate_kbps && *target_bitrate_kbps > 0) {
+        // Mirror the game-stream client-settings path: a live_tuning action
+        // must move the running session, not only the persisted record.
+        adaptive_bitrate::set_base_bitrate(*target_bitrate_kbps);
+      }
       output_tree["status"] = result == nvhttp::client_mutation_result_t::success;
       if (result == nvhttp::client_mutation_result_t::not_found) {
         output_tree["error"] = "Paired client was not found";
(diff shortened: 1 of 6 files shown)


papi-ux:
Closing as superseded. The parked goal here, a web-executable Doctor safe action, landed on master in a different shape while this branch sat: `POST /api/doctor/action` backed by the dedicated `doctor_actions` module, with a full run_id lifecycle including undo (with stale-run guards) and a verify step, plus the evidence-gated one-click flow from #378.

This branch's transport (the scoped client-settings patch with the web supplying the session uuid) and its Dashboard wiring predate that machinery. A rebase attempt confirmed the only conflicts are exactly those two forks: resolving toward this branch regresses the shipped endpoint, resolving toward master leaves an empty diff plus one unused route. The tests are covered too, `StreamStatsDoctorTests` on master asserts the action contract including undo, which these tests never reached. Nothing left to carry over.

Issue #409, opened 2026-08-15
Enable GPU-native capture for VAAPI on the headless private route


papi-ux:
## What

Every VAAPI host captures through a system-memory copy today. `gpu_native_dmabuf_is_safe()` in `src/platform/linux/wlgrab_capture_policy.h` is one line:

```cpp
return hwdevice_type == platf::mem_type_e::cuda;
```

All three capture routes consult it, so direct wlroots desktop, headless ext-image-copy, and windowed-private all refuse the DMA-BUF import for VAAPI before attempting it. NVIDIA hosts stream GPU-native. AMD and Intel hosts pay a full-frame CPU round trip on every frame.

This issue owns the shared predicate and the **headless private route**, which is the one to land first: it has field evidence, and it is the smallest surface, a single headless output whose format we choose. The other two routes are split out and depend on this one:

- #410, windowed-private route
- #411, direct desktop mirror route, deliberately last

This is not a request to flip the flag.

## Why the refusal exists

#144 routed true-headless VAAPI through the conservative SHM path after DMA-BUF import trouble on AMD, alongside #124 (AV1 VA-API slice errors, black client) and #111. #398 extended the same predicate to the direct desktop route, because that path could still hand a tiled physical-monitor modifier to VAAPI, which is the crash being avoided. Both decisions were correct at the time and both are load bearing until something replaces them.

## What it costs, measured

From #367, a single-GPU 7900 XTX CachyOS host streaming 4K AV1 over headless labwc, iGPU disabled in BIOS so there is no cross-device excuse:

- 31.1 fps against a 60 fps target, 0.00% dropped, 0.06% duplicate
- encode 5.6 ms per frame, GPU core at 32%, 107 W
- avg frame age 24.5 ms
- `capture_path=shm_cpu_capture`, bgra8, roughly 33 MB per frame

Nothing downstream is saturated. Capture supply is the whole ceiling, and at 4K the copy costs about a frame time on its own. #92 is the same shape on an RX 9060 XT, where host processing sits near 14 ms against roughly 2 ms on Sunshine for the same host and client.

## Shape of the work

1. Prerequisite hardening, largely landed. #404 validates VAAPI DRM PRIME object, layer, and plane counts before indexing and keeps returned FDs under RAII. #403 made Cage and WLR screencopy requests own their frame metadata and pending protocol objects through timeout and teardown.
[...]

Issue #367, opened 2026-08-11
[Bug]: Stream GPU defaults to CPU/APU instead of GPU


coreydeli:
### Describe the bug

I am streaming to a Rog Ally X running Bazzite w/ Moonlight and a Google Tv Streamer.

Ally X runs at 60FPS though it does request 120
GTV runs at 30FPS requesting 60 wired ethernet on a 10GB backbone network.

Sunshine runs this without issue at near my native desktops FPS but for some reason when I stream using Polaris it does not. It seems like this is trying to use my APU despite my efforts with adding specific markers for the correct GPU card. I always get a SHM fallback as well.

### Steps to reproduce

Stream to google tv

### Install method

Arch package

### Host OS

CachyOS, Polaris 1.3.7

### GPU and driver

AMD 7900XTX Mesa 26.1.6-1

### Client and device

ROG Ally X, Google TV Streamer. Using Moonlight

### Runtime path

Headless

### Relevant log output

```Shell

```

### Additional context

[polaris-anonymized-diagnostics-2026-08-11T14-59-21.015Z.json](https://github.com/user-attachments/files/30942580/polaris-anonymized-diagnostics-2026-08-11T14-59-21.015Z.json)
[polaris-issue-draft-2026-08-11T14-59-19.379Z.md](https://github.com/user-attachments/files/30942579/polaris-issue-draft-2026-08-11T14-59-19.379Z.md)

papi-ux:
Hey there! thanks for the report and especially the diagnostics bundle, always helps!

I dug through the bundle and this is a real dual-GPU bug on the Polaris side, and your APU suspicion is right, so that error is totally on me, I saw that your adapter_name = /dev/dri/renderD128 so that node really is the 7900 XTX (the vaapi vendor probe in your bundle confirms it). But the private compositor reports:

`Wayland DMA-BUF feedback ready: main_device=/dev/dri/renderD129`

That's your APU. In headless mode wlroots picks the first render node it enumerates and adapter_name doesn't reach it, so your games were rendering on the iGPU while only the encode ran on the XTX. Every 4K frame then crossed a CPU copy on the way over, which is where your framerate went. Your encoder itself is totally fine, the bundle shows 5.8 ms encode time at 4K AV1.

[...]

PR #449, opened 2026-08-16
fix(linux): contain VAAPI headless DMA-BUF after v1.3.9 crash

papi-ux:
## Summary

- restore fail-closed GPU-native DMA-BUF eligibility for VAAPI on every wlroots route
- retain the route/modifier plumbing so a hardware-proven guard can replace the containment later
- preserve CUDA GPU-native behavior
- report `vaapi_headless_dmabuf_disabled_for_stability` when the headless VAAPI path falls back to SHM

## Why

An affected AMD/CachyOS host from #367 began dumping `SIGSEGV` at stream start after updating to v1.3.9. The diagnostics bundle was exported after the backend died and contains no crash stack, so the exact faulting function still needs the coredump. The leading regression boundary is the VAAPI headless DMA-BUF import/conversion path enabled by #430 without its affected-host field gate.

This PR contains that path while the crash is diagnosed. A linear modifier is not sufficient evidence that the first live VAAPI import/conversion is safe on the reporter's stack.

## Impact

VAAPI private-headless capture returns to the SHM/system-memory path used before v1.3.9. The 4K performance optimization from #430 is temporarily disabled. CUDA/NVENC GPU-native capture is unchanged, and direct/windowed VAAPI routes remain on their existing SHM containment.

## Validation

- full local build
- focused capture policy and source-contract tests: 9 passed
- cage display router policy tests: 35 passed
- `ctest --test-dir build -j1 --output-on-failure`: 17/17 passed
- `bash scripts/check-public-surface.sh`
- `git diff --check`

## Follow-up

- collect the matching v1.3.9 `coredumpctl` output and final Polaris log from #367
- validate this containment on the affected AMD host
- reopen VAAPI headless GPU-native eligibility only with AMD/Intel first-frame, sustained-cadence, fallback, and crash evidence

Refs #367
Refs #409

### DIFF
--- a/src/platform/linux/cage_display_router.h
+++ b/src/platform/linux/cage_display_router.h
@@ -168,8 +168,8 @@ namespace cage_display_router {
    * @brief Returns whether a headless labwc runtime should try the
    *        ext-image-copy-capture DMA-BUF path before falling back to SHM.
    *
-   * VAAPI may probe only this private headless route. Its captured buffer still
-   * has to pass the modifier policy before it reaches the encoder.
+   * CUDA may probe this private headless route. VAAPI remains fail-closed while
+   * the affected-host crash at its first live import is under containment.
    */
   bool should_attempt_headless_extcopy_dmabuf(
     const platf::runtime_state_t &runtime_state,
--- a/src/platform/linux/wlgrab.cpp
+++ b/src/platform/linux/wlgrab.cpp
@@ -759,6 +759,13 @@ namespace platf {
               "cached_unsupported",
               true
             );
+          } else {
+            stream_stats::update_gpu_native_probe_attempt(
+              "headless_extcopy",
+              "ineligible",
+              "policy",
+              "vaapi_headless_dmabuf_disabled_for_stability"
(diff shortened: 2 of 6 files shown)


papi-ux:
Superseded by merged #447. The #447 head and master merge commit have identical source trees, and the #367 affected-host validation request now targets merged master at 46e6fed3. Closing this duplicate without deleting its branch.

papi-ux:
Superseded without merging.

- The production containment is already on `master` through #447 at merge SHA `46e6fed3b562414af5e3c6041ebb01b1ae0b2fe4`.
- The useful coverage unique to this draft—the all-route/all-modifier VAAPI refusal matrix and retained modifier-guard source contract—now lives in draft #453 at exact head `84c35801e66491d77eb7fb1f012effeb6651a6a5`.
- #453's protected Public Hygiene, Web, C++ sanitizer, and Arch checks are green.

This branch conflicts with the merged implementation, so closing it avoids a second production rewrite while preserving the stronger contracts in a clean test-only PR.

Issue #422, opened 2026-08-15
Nested gamescope for Steam Big Picture breaks keyboard+controller input inside Private Stream


FabianZettl:
### Describe the bug

Follow-up to #400/#401 and to [nova#237](https://github.com/papi-ux/nova/pull/237) (Ctrl+1/Ctrl+2 quick keys for Big Picture Menu / Quick Access Menu). Trying to get a full Steam Deck-like in-game overlay experience over a Polaris Private Stream session surfaced three separate rough edges, filing them together since they're all part of the same "Steam Big Picture through Polaris" story.

**1. Wrapping the "Steam Big Picture" app in a nested gamescope session breaks keyboard *and* controller input**

To get gamescope's `--steam` IPC bridge (needed for Ctrl+1/Ctrl+2 to reach Steam while a game has focus — see nova#237), I pointed the "Steam Big Picture" app's `detached` command at a wrapper script that launches gamescope itself:

```bash
gamescope --steam -W "$width" -H "$height" -w "$width" -h "$height" -- steam -gamepadui
```

...launched as an ordinary app inside an already-running `headless_stream` (Private Stream / labwc) session — i.e. gamescope ends up as a nested Wayland client of Polaris' private labwc compositor, not as Polaris' own owned `gamescope_runtime`.

gamescope starts fine and Steam runs inside it (confirmed via `ps`: `gamescope --steam -W 1920 -H 1080 -w 1920 -h 1080 -- steam -gamepadui`), but once a game launches from within it, **no keyboard input and no controller input reach the game** at all. Not just Ctrl+1/2 — the entire More Keys sheet (ESC, Alt+F4, etc.) and the controller stop working. Reverting the app's `detached` command back to plain `setsid steam -gamepadui` (no gamescope) immediately restores both.

**2. `polaris-gamescope-session` isn't usable outside the `gamescope_stream` architecture**

[...]

PR #481, opened 2026-08-18
fix(linux): keep PipeWire VAAPI capture safe by default

### LINKED ISSUE #480: gamescope_stream: AMD/VAAPI gamescopegrab stalls after first captured frame
## Release gate

This is split from the latest affected-host report on #422 and is part of the open gamescope/PipeWire productization gate in #152.

- Field report: https://github.com/papi-ux/polaris/issues/422#issuecomment-5326544071
- Reported build: v1.3.10, `daccc762a53a28c4df2b7da81bc0806ad50dbb34`
- Affected stack: CachyOS/Hyprland, AMD Radeon RX 7800 XT, Mesa 26.1.6, Gamescope 3.16.24, Nova on AYN Thor
- Mode: `linux_stream_mode=gamescope_stream`, `capture=portal`, HEVC VAAPI
- v1.3.11 candidate: `a91e4345fa21b2b6e603007318218fdefe427c64`

The relevant `portal_grab.cpp`, `pipewire_capture.cpp`, and `portal_session.cpp` code is unchanged between the reported v1.3.10 commit and the v1.3.11 candidate. Keep Polaris v1.3.11 and the matched Nova release draft until this gate is cleared.

## Observed behavior

The regular xdg-desktop-portal attempt still times out. Polaris then discovers Gamescope's local PipeWire `Video/Source` and negotiates a GPU-native DMA-BUF path:

```text
pipewire: found gamescope Video/Source ...
portal: gamescopegrab local Video/Source ...
portal: PipeWire DMA-BUF path enabled ...
portal: PipeWire format negotiated: 1920x1080 ...
Creating encoder [hevc_vaapi]
Using frame converter [avcodec_encode_device] target_device=vaapi target_residency=gpu
portal: alloc_img 1920x1080 ...
portal: capture_transport=dmabuf frame_residency=gpu frame_format=bgra8
portal: alloc_img 1920x1080 ...
```

It then becomes silent. Nova remains black, no media packets reach the client, Polaris consumes no advancing CPU time, no timeout or clean shutdown occurs, and the process requires `kill -9`.

This is not being classified as a v1.3.10 regression without an older good boundary.

## Important diagnostic boundary

Polaris emits `capture_transport=dmabuf` only after `wait_for_frame()` and `fill_frame()` succeed. At least one DMA-BUF frame therefore reached the capture-to-encode handoff. The exact stall may still be in VAAPI import/conversion, encoder submission, buffer lifetime, or the next producer frame; do not name the faulting component without a thread dump or equivalent evidence.

The v1.3.11 RP6 smoke used the NVIDIA `headless_stream` ext-image-copy path, not AMD/VAAPI gamescopegrab, so it does not clear this failure.

## Required reproduction

On the affected host, test the exact v1.3.11 candidate twice:

1. Default DMA-BUF behavior.
2. `POLARIS_PORTAL_DMABUF=0`.

For each run retain privacy-safe evidence for:

- gamescope node id/serial and negotiated PipeWire format/data type
- `capture_transport`, `frame_residency`, encoder API, and conversion target
- whether the client receives changing frames and media packets
- a thread dump or wait-channel snapshot during any stall
- session stop/cleanup and `coredumpctl`

The forced-SHM run must be checked for an actual `capture_transport=shm frame_residency=cpu` result. If the local gamescope node keeps negotiating DMA-BUF despite the escape hatch, that is part of this bug.

## Acceptance criteria

- A missing first encoded packet or continuing frame flow cannot leave the session black indefinitely.
- DMA-BUF refusal, import failure, or stalled handoff falls back to a working SHM path, or terminates the session with a clear actionable error if safe fallback is impossible.
- Forced SHM streams continuously on the affected host and reports SHM/CPU residency truthfully.
- Default mode produces changing frames and media packets, or performs the documented bounded fallback.
- End Session and failure cleanup leave no Polaris/Gamescope/session-marker/media-socket residue.
- No new Polaris coredump.
- Add regression coverage for the timeout/fallback policy and the forced-SHM offer path.

papi-ux:
## What

- Fail closed to SHM before PipeWire format advertisement for VAAPI capture on both local gamescope/KWin graphs and portal-remoted streams.
- Preserve CUDA DMA-BUF behavior.
- Treat `POLARIS_PORTAL_DMABUF` as a narrow typed override:
  - unset or any value other than exact `0`/`1`: safe default, so VAAPI stays on SHM
  - exact `0`: force SHM, including a second check at final PipeWire negotiation
  - exact `1`: explicitly opt VAAPI back into DMA-BUF for field-proven hosts
- Keep the opt-in behind the existing local-graph LINEAR route and the portal-remote canonical render-node/EGL modifier checks.
- Emit a warning that the opt-in is unvalidated and has no automatic stall fallback.

## Why

The affected RX 7800 XT report on v1.3.10 negotiates `capture_transport=dmabuf frame_residency=gpu`, then produces a black/no-media session that requires `kill -9`. The trace proves that one frame reached the capture-to-encode handoff; it does not distinguish a later import, conversion, encode, buffer-lifetime, or producer stall.

The old eligibility check proved matching render nodes and an importable layout, but that is not proof of driver-safe VAAPI import/conversion lifetime. Defaulting VAAPI to SHM is the bounded containment.

A second AMD host also proved that a permanent blanket block would be too broad: RX 9070 XT / RDNA4 / Mesa 26.1.6 / KWin streamed correctly through the unpatched local DMA-BUF path, while the same hardware still hit the separate zero-frame failure on headless ext-image-copy. Exact `POLARIS_PORTAL_DMABUF=1` therefore exposes the already-existing PipeWire VAAPI route only to operators who have proved it on their host; it is not an allowlist or a claim that AMD VAAPI is generally safe.

A generic no-frame timeout remains intentionally out of scope: a static Gamescope scene can legitimately stop producing frames, so a watchdog would create false reconnect loops.

## Validation

RED before the opt-in implementation:

- with the final tests kept and only the production files restored to the original #481 source, the build failed because `dmabuf_override_e`, its exact-value parser, and the typed policy overload did not exist

GREEN on Linux after the implementation:

- focused default/force/opt-in policy and source-boundary tests
- exact environment parser counter-tests (`00`, `01`, `true`, whitespace, unset)
- `test_polaris`: 353/353
- `test_polaris_platform`: 218 passed, 1 expected CUDA-only skip
- full production `polaris` executable build
- `git diff --check`

Protected exact-head CI is required after this rebase.

## Release gate

Keep this PR draft until the affected RX 7800 XT / VAAPI host proves on the exact head:

- default `gamescope_stream` produces changing media through the SHM containment
- exact `POLARIS_PORTAL_DMABUF=0` behaves the same way from initial negotiation
- session stop/cleanup completes without a coredump, residue, or forced kill

The RDNA4 host may separately test exact `POLARIS_PORTAL_DMABUF=1`; that is useful opt-in evidence but does not replace the affected-host default-containment gate.

Closes #480. Refs #409. Original affected report: https://github.com/papi-ux/polaris/issues/422#issuecomment-5326544071. RDNA4 local-graph evidence: https://github.com/papi-ux/polaris/pull/481#issuecomment-5334129897.

### DIFF
--- a/src/platform/linux/pipewire_capture.cpp
+++ b/src/platform/linux/pipewire_capture.cpp
@@ -78,10 +78,6 @@ namespace pipewire_capture {
              spa_format == SPA_VIDEO_FORMAT_RGBA;
     }
 
-    bool gpu_encoder_mem_type(platf::mem_type_e mem_type) {
-      return mem_type == platf::mem_type_e::vaapi || mem_type == platf::mem_type_e::cuda;
-    }
-
     bool contains_modifier(const std::vector<std::uint64_t> &modifiers, std::uint64_t modifier) {
       return std::find(modifiers.begin(), modifiers.end(), modifier) != modifiers.end();
     }
@@ -147,7 +143,28 @@ namespace pipewire_capture {
     return canonical_render_node(nodes.front());
   }
 
-  bool may_offer_dmabuf(const dmabuf_eligibility_t &eligibility) {
+  dmabuf_override_e dmabuf_override_from_env(const char *value) {
+    if (value && std::strcmp(value, "0") == 0) {
+      return dmabuf_override_e::force_cpu;
+    }
+    if (value && std::strcmp(value, "1") == 0) {
+      return dmabuf_override_e::allow_vaapi;
+    }
+    return dmabuf_override_e::default_safe;
+  }
+
+  bool may_offer_dmabuf(const dmabuf_eligibility_t &eligibility, dmabuf_override_e override) {
+    // #367 and #480 reached the first-frame VAAPI failure boundary on different
+    // capture backends. A matching render node and importable modifier prove
+    // layout compatibility, not driver-safe import/conversion lifetime. Keep
+    // every unproven VAAPI PipeWire route on SHM by default. CUDA remains the
+    // validated GPU-native path; VAAPI is available only as an explicit field-
+    // testing opt-in on hosts where the operator has already proved it works.
+    const bool supported_encoder = eligibility.mem_type == platf::mem_type_e::cuda ||
+                                   (override == dmabuf_override_e::allow_vaapi && eligibility.mem_type == platf::mem_type_e::vaapi);
+    if (override == dmabuf_override_e::force_cpu || !supported_encoder) {
+      return false;
+    }
     if (!eligibility.capture_render_node) {
       return false;
     }
@@ -155,7 +172,6 @@ namespace pipewire_capture {
     const auto encoder_node = canonical_render_node(eligibility.encoder_render_node);
     return capture_node && encoder_node &&
            *capture_node == *encoder_node &&
-           gpu_encoder_mem_type(eligibility.mem_type) &&
            eligibility.egl_import_supported;
   }
 
@@ -1064,8 +1080,8 @@ namespace pipewire_capture {
       cap->set_terminal(wait_result_e::reinit);
       return;
     }
-    const char *dmabuf_env = std::getenv("POLARIS_PORTAL_DMABUF");
-    const bool force_cpu = dmabuf_env && dmabuf_env[0] == '0' && dmabuf_env[1] == '\0';
+    const auto dmabuf_override = dmabuf_override_from_env(std::getenv("POLARIS_PORTAL_DMABUF"));
+    const bool force_cpu = dmabuf_override == dmabuf_override_e::force_cpu;
     const bool want_dmabuf = cap->options_.may_use_dmabuf && !force_cpu;
     const auto has_modifier = (raw_info.flags & SPA_VIDEO_FLAG_MODIFIER) &&
                               raw_info.modifier != DRM_FORMAT_MOD_INVALID;
--- a/src/platform/linux/pipewire_capture.h
+++ b/src/platform/linux/pipewire_capture.h
@@ -77,6 +77,12 @@ namespace pipewire_capture {
     bool egl_import_supported = false;
   };
 
+  enum class dmabuf_override_e {
+    default_safe,
+    force_cpu,
+    allow_vaapi,
+  };
+
   struct dmabuf_plane_t {
     int fd = -1;
     std::uint32_t chunk_offset = 0;
@@ -111,7 +117,16 @@ namespace pipewire_capture {
     const std::optional<std::string> &portal_capture_render_node,
     const std::optional<std::string> &encoder_render_node
   );
-  bool may_offer_dmabuf(const dmabuf_eligibility_t &eligibility);
+  // Only exact "0" and "1" values carry authority. Everything else, including
+  // a missing variable, retains the fail-closed default.
+  dmabuf_override_e dmabuf_override_from_env(const char *value);
+  // VAAPI stays fail-closed unless the operator explicitly opts into the
+  // unvalidated route. The typed override prevents contradictory force/allow
+  // flags from reaching this boundary.
+  bool may_offer_dmabuf(
+    const dmabuf_eligibility_t &eligibility,
+    dmabuf_override_e override = dmabuf_override_e::default_safe
+  );
   std::vector<dmabuf_format_modifier_t> task1_packed_dmabuf_formats(std::vector<std::uint64_t> modifiers);
   std::vector<dmabuf_format_modifier_t> filter_importable_dmabuf_formats(
     const std::vector<dmabuf_format_modifier_t> &portal_formats,
--- a/src/platform/linux/portal_grab.cpp
+++ b/src/platform/linux/portal_grab.cpp
@@ -178,6 +178,29 @@ namespace portal {
     return client_dynamic_range <= 0;
   }
 
+  static pipewire_capture::dmabuf_override_e portal_dmabuf_override() {
+    return pipewire_capture::dmabuf_override_from_env(std::getenv("POLARIS_PORTAL_DMABUF"));
+  }
+
+  static void log_dmabuf_policy(
+    platf::mem_type_e mem_type,
+    pipewire_capture::dmabuf_override_e override,
+    std::string_view source
+  ) {
+    if (override == pipewire_capture::dmabuf_override_e::force_cpu) {
+      BOOST_LOG(info) << "portal: portal_dmabuf_forced_off source="sv << source
+                      << "; offering SHM only"sv;
+    }
+    else if (mem_type == platf::mem_type_e::vaapi && override == pipewire_capture::dmabuf_override_e::allow_vaapi) {
+      BOOST_LOG(warning) << "portal: vaapi_pipewire_dmabuf_explicitly_enabled source="sv << source
+                         << "; operator opted into an unvalidated path with no automatic stall fallback"sv;
+    }
+    else if (mem_type == platf::mem_type_e::vaapi) {
+      BOOST_LOG(info) << "portal: vaapi_pipewire_dmabuf_disabled_for_stability source="sv << source
+                      << "; offering SHM by default; set POLARIS_PORTAL_DMABUF=1 only on a host where this path is known to work"sv;
+    }
+  }
+
   // Encoder render node for DMA-BUF eligibility: the configured adapter_name
   // when it names a canonical render node, else the host's sole render node.
   // Single-GPU hosts get zero-copy without configuration; multi-GPU hosts stay
@@ -219,7 +242,18 @@ namespace portal {
     bool may_use_dmabuf = false;
     const bool prefer_hdr = portal_prefer_hdr_formats(client_dynamic_range);
     const bool prefer_sdr = portal_prefer_sdr_formats(client_dynamic_range);
+    const auto dmabuf_override = portal_dmabuf_override();
     if (encoder_render_node) {
+      const pipewire_capture::dmabuf_eligibility_t eligibility {
+        .capture_render_node = encoder_render_node,
+        .encoder_render_node = *encoder_render_node,
+        .mem_type = mem_type,
+        .egl_import_supported = true,
+      };
+      may_use_dmabuf = pipewire_capture::may_offer_dmabuf(eligibility, dmabuf_override);
+    }
+    log_dmabuf_policy(mem_type, dmabuf_override, "local_graph"sv);
+    if (may_use_dmabuf) {
       // LINEAR always for gamescope (only allocates LINEAR on PW node).
       dmabuf_formats = pipewire_capture::task1_packed_dmabuf_formats({DRM_FORMAT_MOD_LINEAR});
       // Ensure 10-bit LINEAR is present for HDR streams even if EGL skipped it.
@@ -247,7 +281,6 @@ namespace portal {
                  format.spa_format == SPA_VIDEO_FORMAT_xRGB_210LE;
         });
       }
-      may_use_dmabuf = !dmabuf_formats.empty();
     }
     auto local = std::make_shared<pipewire_capture::capture_t>(pipewire_capture::capture_options_t {
       .remote_fd = -1,
@@ -568,15 +601,23 @@ namespace portal {
         }
         std::vector<pipewire_capture::dmabuf_format_modifier_t> dmabuf_formats;
         bool may_use_dmabuf = false;
-        if (!session->capture_render_node) {
+        const auto dmabuf_override = portal_dmabuf_override();
+        const bool allow_vaapi = dmabuf_override == pipewire_capture::dmabuf_override_e::allow_vaapi;
+        log_dmabuf_policy(mem_type, dmabuf_override, "portal_remote"sv);
+        if (dmabuf_override == pipewire_capture::dmabuf_override_e::force_cpu ||
+            (mem_type == platf::mem_type_e::vaapi && !allow_vaapi)) {
+          // The policy log above records whether this is an operator-forced or
+          // default stability containment decision.
+        } else if (!session->capture_render_node) {
           BOOST_LOG(info) << "portal: DMA-BUF disabled because the portal stream did not provide an explicit capture render node"sv;
         } else if (!encoder_render_node) {
           BOOST_LOG(info) << "portal: DMA-BUF disabled because config::video.adapter_name is not an explicit canonical render node"sv;
         } else if (*session->capture_render_node != *encoder_render_node) {
           BOOST_LOG(info) << "portal: DMA-BUF disabled because capture render node ["sv << *session->capture_render_node
                           << "] does not match encoder adapter ["sv << *encoder_render_node << ']';
-        } else if (mem_type != platf::mem_type_e::vaapi && mem_type != platf::mem_type_e::cuda) {
-          BOOST_LOG(info) << "portal: DMA-BUF disabled because encoder memory type is not VAAPI or CUDA"sv;
+        } else if (mem_type != platf::mem_type_e::cuda &&
+                   !(allow_vaapi && mem_type == platf::mem_type_e::vaapi)) {
+          BOOST_LOG(info) << "portal: DMA-BUF disabled because encoder memory type is neither CUDA nor explicitly enabled VAAPI"sv;
         } else {
           if (mem_type == platf::mem_type_e::cuda) {
             // LINEAR one-plane packed RGB: 8-bit BGRx/BGRA + 10-bit xBGR_210LE (HDR).
@@ -633,7 +674,7 @@ namespace portal {
             .mem_type = mem_type,
             .egl_import_supported = !dmabuf_formats.empty(),
           };
-          may_use_dmabuf = pipewire_capture::may_offer_dmabuf(eligibility);
+          may_use_dmabuf = pipewire_capture::may_offer_dmabuf(eligibility, dmabuf_override);
           if (!may_use_dmabuf) {
             BOOST_LOG(info) << "portal: DMA-BUF disabled because no compatible packed-RGB modifier is available for encoder render node ["sv
                             << *encoder_render_node << ']';
@@ -646,6 +687,7 @@ namespace portal {
                         << " prefer_hdr="sv << (want_prefer_hdr ? "true"sv : "false"sv)
                         << " prefer_sdr="sv << (want_prefer_sdr ? "true"sv : "false"sv)
                         << " force_hdr="sv << (portal_force_hdr_enabled() ? "true"sv : "false"sv)
+                        << " vaapi_opt_in="sv << (allow_vaapi ? "true"sv : "false"sv)
                         << " client_dynamic_range="sv << client_dynamic_range
                         << " capture_node="sv << session->capture_render_node.value_or("none")
                         << " encoder_node="sv << encoder_render_node.value_or("none")
--- a/tests/unit/platform/test_portal_grab_policy.cpp
+++ b/tests/unit/platform/test_portal_grab_policy.cpp
@@ -628,15 +628,34 @@ TEST(PipeWireCapturePolicyTests, RenderNodeValidationAcceptsOnlyCanonicalRenderN
   EXPECT_EQ(pipewire_capture::canonical_render_node("/dev/dri/renderDabc"), std::nullopt);
 }
 
-TEST(PipeWireCapturePolicyTests, SameRenderNodeEligibilityRequiresExplicitMatchingGpuPathAndEglSupport) {
+TEST(PipeWireCapturePolicyTests, DmaBufEligibilityRequiresSupportedEncoderAndExplicitMatchingGpuPath) {
   const pipewire_capture::dmabuf_eligibility_t eligible {
     .capture_render_node = "/dev/dri/renderD128",
     .encoder_render_node = "/dev/dri/renderD128",
-    .mem_type = platf::mem_type_e::vaapi,
+    .mem_type = platf::mem_type_e::cuda,
     .egl_import_supported = true,
   };
 
   EXPECT_TRUE(pipewire_capture::may_offer_dmabuf(eligible));
+  EXPECT_FALSE(pipewire_capture::may_offer_dmabuf(eligible, pipewire_capture::dmabuf_override_e::force_cpu));
+
+  auto vaapi = eligible;
+  vaapi.mem_type = platf::mem_type_e::vaapi;
+  EXPECT_FALSE(pipewire_capture::may_offer_dmabuf(vaapi));
+  EXPECT_TRUE(pipewire_capture::may_offer_dmabuf(vaapi, pipewire_capture::dmabuf_override_e::allow_vaapi));
+  EXPECT_FALSE(pipewire_capture::may_offer_dmabuf(vaapi, pipewire_capture::dmabuf_override_e::force_cpu));
+
+  auto vaapi_missing_capture = vaapi;
+  vaapi_missing_capture.capture_render_node.reset();
+  EXPECT_FALSE(pipewire_capture::may_offer_dmabuf(vaapi_missing_capture, pipewire_capture::dmabuf_override_e::allow_vaapi));
+
+  auto vaapi_mismatched = vaapi;
+  vaapi_mismatched.encoder_render_node = "/dev/dri/renderD129";
+  EXPECT_FALSE(pipewire_capture::may_offer_dmabuf(vaapi_mismatched, pipewire_capture::dmabuf_override_e::allow_vaapi));
+
+  auto vaapi_no_egl = vaapi;
+  vaapi_no_egl.egl_import_supported = false;
+  EXPECT_FALSE(pipewire_capture::may_offer_dmabuf(vaapi_no_egl, pipewire_capture::dmabuf_override_e::allow_vaapi));
 
   auto missing_capture = eligible;
   missing_capture.capture_render_node.reset();
@@ -649,6 +668,7 @@ TEST(PipeWireCapturePolicyTests, SameRenderNodeEligibilityRequiresExplicitMatchi
   auto system_memory = eligible;
   system_memory.mem_type = platf::mem_type_e::system;
   EXPECT_FALSE(pipewire_capture::may_offer_dmabuf(system_memory));
+  EXPECT_FALSE(pipewire_capture::may_offer_dmabuf(system_memory, pipewire_capture::dmabuf_override_e::allow_vaapi));
 
   auto no_egl = eligible;
   no_egl.egl_import_supported = false;
@@ -658,6 +678,19 @@ TEST(PipeWireCapturePolicyTests, SameRenderNodeEligibilityRequiresExplicitMatchi
   noncanonical.capture_render_node = "renderD128";
   noncanonical.encoder_render_node = "renderD128";
   EXPECT_FALSE(pipewire_capture::may_offer_dmabuf(noncanonical));
+}
+
+TEST(PipeWireCapturePolicyTests, DmaBufEnvironmentOverrideRequiresAnExactZeroOrOne) {
+  using enum pipewire_capture::dmabuf_override_e;
+
+  EXPECT_EQ(pipewire_capture::dmabuf_override_from_env(nullptr), default_safe);
+  EXPECT_EQ(pipewire_capture::dmabuf_override_from_env(""), default_safe);
+  EXPECT_EQ(pipewire_capture::dmabuf_override_from_env("0"), force_cpu);
+  EXPECT_EQ(pipewire_capture::dmabuf_override_from_env("1"), allow_vaapi);
+  EXPECT_EQ(pipewire_capture::dmabuf_override_from_env("00"), default_safe);
+  EXPECT_EQ(pipewire_capture::dmabuf_override_from_env("01"), default_safe);
+  EXPECT_EQ(pipewire_capture::dmabuf_override_from_env("true"), default_safe);
+  EXPECT_EQ(pipewire_capture::dmabuf_override_from_env(" 1"), default_safe);
 }
 
 TEST(PipeWireCapturePolicyTests, DmaBufCapabilityFilteringKeepsOnlyPackedRgbImportableNonExternalFormats) {
--- a/tests/unit/test_linux_platform_hardening_source.cpp
+++ b/tests/unit/test_linux_platform_hardening_source.cpp
@@ -50,6 +50,26 @@ TEST(LinuxPlatformHardeningSource, HeadlessVaapiFailsClosedBeforeDmabufInitializ
   EXPECT_NE(process.find("Retaining headless extcopy failure reason"), std::string::npos);
 }
 
+TEST(LinuxPlatformHardeningSource, PipeWireVaapiDefaultsToShmWithExplicitOptInAndForcedCpuPrecedence) {
+  const auto policy = read_source("src/platform/linux/pipewire_capture.cpp");
+  const auto capture = read_source("src/platform/linux/portal_grab.cpp");
+
+  EXPECT_NE(policy.find("override == dmabuf_override_e::allow_vaapi && eligibility.mem_type == platf::mem_type_e::vaapi"), std::string::npos);
+  EXPECT_NE(policy.find("dmabuf_override_from_env(std::getenv(\"POLARIS_PORTAL_DMABUF\"))"), std::string::npos);
+  EXPECT_NE(capture.find("portal_dmabuf_override"), std::string::npos);
+  EXPECT_NE(capture.find("vaapi_pipewire_dmabuf_disabled_for_stability"), std::string::npos);
+  EXPECT_NE(capture.find("vaapi_pipewire_dmabuf_explicitly_enabled"), std::string::npos);
+
+  const std::string offer_call = "may_offer_dmabuf(eligibility, dmabuf_override)";
+  const auto local_offer = capture.find(offer_call);
+  ASSERT_NE(local_offer, std::string::npos);
+  EXPECT_NE(capture.find(offer_call, local_offer + offer_call.size()), std::string::npos)
+    << "both local-graph and portal-remote capture must apply the same explicit policy";
+
+  EXPECT_NE(capture.find("query_egl_dmabuf_import_formats(*encoder_render_node)"), std::string::npos)
+    << "the explicit VAAPI opt-in must retain the existing EGL modifier capability gate";
+}
+
 TEST(LinuxPlatformHardeningSource, HeadlessModifierGuardRemainsBehindVaapiContainment) {
   const auto header = read_source("src/platform/linux/wayland.h");
   const auto capture = read_source("src/platform/linux/wlgrab.cpp");


CarlosSouza:
Affected-host report for the release gate, from a second AMD/VAAPI host.

**The headline is awkward for the PR: this host does not reproduce #480. Unpatched VAAPI DMA-BUF capture works here and produces good video.** The containment itself passes all three gate items cleanly, so the gate is satisfiable — but on this host it closes off a path that was already working.

## Host

| | |
|---|---|
| GPU | RX 9070 XT (Navi 48, **gfx1201**, RDNA4) — single GPU, iGPU off in firmware, one render node |
| Driver | Mesa 26.1.6-arch3.1 (radeonsi / RADV, ACO), DRM 3.64, VA-API 1.24.0 |
| OS | CachyOS, kernel 7.1.8-1-cachyos, **KDE Plasma 6.7 Wayland** |
| gamescope | **3.16.24** (same version as the #480 report) |
| Mode | `linux_stream_mode = gamescope_stream`, `encoder = vaapi`, capture left unset |
| Client | VoidLink on iPad Mini 7, 1920x1080, HDR off |

Differences from the #480 reporter: RDNA4 vs RDNA3, KWin vs Hyprland. Same Mesa, same gamescope.

Built from source at `279b358` (PR head) and at its parent `ddf8d21` for the baseline, same tree, same flags. Verified by string inspection that only the PR binary carries `vaapi_pipewire_dmabuf_disabled_for_stability` / `portal_dmabuf_forced_off`, and only the base retains `memory type is not VAAPI or CUDA`.

## Gate results

**1. Default `gamescope_stream` produces media through the SHM containment — PASS**

```
portal: vaapi_pipewire_dmabuf_disabled_for_stability source=local_graph
portal: dmabuf_negotiate may_use=false force_cpu=false has_modifier=false modifier=0 spa_format=8 formats=0 offered=false negotiated=false flags=0
portal: PipeWire format negotiated: 1920x1080 format=Spa:Enum:VideoFormat:BGRx
portal: capture_transport=shm frame_residency=cpu frame_format=bgra8
```

DMA-BUF was never offered, so the containment lands before format advertisement as intended. `h264_vaapi` and `hevc_vaapi` were the session encoders. Steam Big Picture streamed with working video and audio.

**2. Forced `POLARIS_PORTAL_DMABUF=0` behaves the same from initial negotiation — PASS**

```
portal: portal_dmabuf_forced_off source=local_graph
portal: dmabuf_negotiate may_use=false force_cpu=true has_modifier=false modifier=0 spa_format=8 formats=0 offered=false negotiated=false flags=0
portal: capture_transport=shm frame_residency=cpu frame_format=bgra8
```

Byte-identical outcome to the default path apart from the reason marker.

**3. Stop/cleanup without coredump or forced kill — PASS**

| Run | teardown | SIGKILL needed | new coredumps |
|---|---|---|---|
| base `ddf8d21` | 4.0s | no | 0 |
| PR default | 3.5s | no | 0 |
| PR `POLARIS_PORTAL_DMABUF=0` | 3.5s | no | 0 |

SIGTERM to the Polaris process, 15s allowance before escalating. No escalation was ever needed.

## The baseline result

On the unpatched parent commit, the same client and app:

```
portal: dmabuf_negotiate may_use=true force_cpu=false has_modifier=true modifier=0 spa_format=8 formats=4 offered=true negotiated=true flags=4
portal: PipeWire DMA-BUF path enabled (modifier=0)
portal: capture_transport=dmabuf frame_residency=gpu frame_format=bgra8
```

Video and audio both worked. No black frame, no first-frame failure, no `kill -9`, no coredump. The failure this PR contains does not occur on RDNA4 + Mesa 26.1.6 + KWin.

So the fail-closed rule is correct as *containment* but is not driver-wide on AMD: at minimum it is not RDNA4-wide on this Mesa. If the eventual goal is a predicate rather than a blanket, this host is a data point for the allow side, and I am happy to run any discriminating test you want.

## What I could not measure

No `capture_telemetry` line in any run, so **no latency comparison between the two transports**. It needs `CAPTURE_PROFILE_SUMMARY_FRAMES = 600` consecutive frames, and a gamescope scene only produces frames on damage. Steam Big Picture sits static, so the counter never got close.

I caught single spot readings from the client of ~5ms on the DMA-BUF baseline and ~1.9ms on the SHM PR build, but they were taken during different momentary animations with different content on screen. **They are not comparable and should not be read as SHM being faster.** I did not want to hand you a number I could not stand behind.

Incidentally this is field confirmation of the reasoning in your PR description for not adding a no-frame watchdog. With a static scene the client reports `-/-/-` host processing latency indefinitely; a Steam friend-login notification animating on screen made real numbers appear for a few seconds, then it returned to `-/-/-`. A watchdog really would fire on a legitimately idle scene.

## Three unrelated bugs found while running this

All on the `gamescope_stream` path, present in both binaries, none caused by this PR. Happy to split these into their own issues if useful.

**a. The gamescope runtime outlives Polaris, and the next session attaches to the dead one.** Teardown logs:

```
process: exact Gamescope root is outside the authorized freeze set pid=<n>
process: refusing all later teardown because exact Gamescope ancestry closure was incomplete
process: retaining immutable cage generation because exact-generation cleanup was incomplete
```

and the process survives Polaris exiting entirely. I confirmed one running 6 minutes after its Polaris had gone, having survived a full process exit. The next Polaris then logs:

```
gamescope_runtime: attached to validated gamescope-0 owner pid=<pid from the dead session>
```

captures that runtime's still-registered PipeWire node, which has nothing rendering into it, and **streams black**. This is the #483 family but a different mechanism, and #485 explicitly leaves private runtime lifetime unchanged, so it is not covered.

**b. Draining the runtime leaves its socket, which then locks out `gamescope_stream` permanently.** On reconnect within one Polaris process:

```
gamescope_runtime: drained owned gamescope group=<n>
gamescope_runtime: refusing unmarked/unowned gamescope-0 socket
gamescope_runtime: gamescope-0 exists but is not a validated owner; refusing destructive cleanup
session_manager: Failed to start cage compositor
```

Every later session fails with `failed to initialize video capture/encoding` until someone manually deletes `$XDG_RUNTIME_DIR/gamescope-0*`. No message names the socket or the directory. Reproduced on demand: first connect fails one way, every retry fails this way.

Taken together (a) and (b) are a bad pair — the two teardown paths fail in opposite directions, one leaving the process alive and one leaving the socket behind.

**c. `polaris-gamescope-session` is not on PATH on Arch**, so every session takes the attach path:

```
session_manager: cannot nest <app> because polaris-gamescope-session is not on PATH; using attach
session_launch: gamescope attach env DISPLAY=:1 WAYLAND_DISPLAY=gamescope-0 ...
```

The script exists only at `nix/modules/polaris-gamescope-session.sh` and is not in the Arch package (`pacman -Ql polaris` has no gamescope entry). The user-visible consequence is that **client input drives the host desktop instead of the streamed session** — the mouse cursor moves on the host monitor. That also makes the private session impossible to interact with, which is what stopped me producing sustained motion for the telemetry above.

## Method note

Harness stops Sunshine for the port conflict and restores it afterwards, clears `encoder_cache.txt` before each run, and reaps leaked gamescope runtimes plus stale sockets between runs. Two runs were discarded before I trusted any of this: one where a stale `software` entry in the encoder cache meant the session ran on `libx264` and proved nothing about VAAPI, and one where the session had silently attached to an orphaned runtime per (a). The harness now asserts a VAAPI encoder was actually used and refuses to report a run otherwise.

papi-ux:
@CarlosSouza, the RDNA4/KWin result is now represented as a narrow escape hatch rather than an allowlist. At exact head `5b72a3bc595a94e2a63d0c809cde8310782aac7e`, VAAPI PipeWire capture defaults to SHM, exact `0` forces SHM, and exact `POLARIS_PORTAL_DMABUF=1` opts back into the prior DMA-BUF path.

If you are willing, one `gamescope_stream` run with exact `1` on the same proven-good RX 9070 XT host would be useful. Expected evidence is:

- `vaapi_pipewire_dmabuf_explicitly_enabled source=local_graph`
- `dmabuf_negotiate may_use=true ... negotiated=true`
- `capture_transport=dmabuf frame_residency=gpu`
- changing video and working audio/media for at least 60 seconds
- normal stop with no forced kill, coredump, owned-runtime residue, or `gamescope-0*` socket residue

The warning intentionally says there is no automatic stall fallback. This opt-in applies only to the PipeWire local-graph/portal routes in #481; it does not lift the separate headless ext-image-copy VAAPI guard discussed in #409.

CarlosSouza:
Ran the exact-`1` opt-in twice on the same RX 9070 XT host, at head `5b72a3b`. **All four log/teardown expectations you listed pass; the fifth (no socket residue) fails in both runs.** Two new `gamescope_stream` problems showed up, and one of your merged fixes is confirmed working here.

Host unchanged from the previous report: RX 9070 XT (gfx1201, RDNA4), Mesa 26.1.6-arch3.1, kernel 7.1.8-1-cachyos, KDE Plasma 6.7 Wayland, gamescope 3.16.24, VoidLink on iPad Mini 7. `linux_stream_mode = gamescope_stream`, `encoder = vaapi`, capture unset, HDR off. Built from source at `5b72a3b`; verified the binary carries `vaapi_pipewire_dmabuf_explicitly_enabled` and that the parser accepts only exact `0`/`1`.

## Your expected evidence

Identical in both runs:

```
portal: vaapi_pipewire_dmabuf_explicitly_enabled source=local_graph
portal: dmabuf_negotiate may_use=true force_cpu=false has_modifier=true modifier=0 spa_format=8 formats=4 offered=true negotiated=true flags=4
portal: capture_transport=dmabuf frame_residency=gpu frame_format=bgra8
```

Session encoders were `h264_vaapi` + `hevc_vaapi`. Sonic Racing CrossWorlds ran at 3840x2160@60 with changing video and working audio well past 60 seconds — run 1 held a session for 105s, run 2 longer and with the game actually playable.

| | run 1 | run 2 |
|---|---|---|
| teardown | 3.5s | 11.0s |
| forced SIGKILL | no | no |
| new coredumps | 0 | 0 |
| owned-runtime residue | **no** | **no** |
| `gamescope-0*` socket residue | **6 files** | **5 files** |

**#493 is confirmed working on this host.** The orphaned owned runtime I reported from the base is gone — no surviving generation, and no later session attaching to a dead marker.

## Socket residue still present

Run 1 left `gamescope-0`, `gamescope-0-ei`, `gamescope-0.lock`, `gamescope-0-ei.lock` and two `gamescope-limiter-XX*`. Run 2 reclaimed `gamescope-0` but left `gamescope-0-ei`, both `.lock` files and two limiters. The `-ei` half is inside the pair #493 says it reclaims; the `.lock` and `gamescope-limiter-*` files look outside its stated scope.

In run 1 Polaris did invoke the documented hook and it failed:

```
Skipping Steam -shutdown undo; polaris-gamescope-session stop owns nested Steam cleanup
Executing Undo Cmd: [.../polaris-gamescope-session stop]
session_manager: waiting up to 30s for nested gamescope teardown
Warning: Return code [1]
```

**Caveat, and I cannot resolve it from here:** a source build does not install `polaris-gamescope-session`, so I staged it and `polaris-gamescope-runtime-lib.sh` by hand out of `nix/modules/` onto `PATH`. Running the same script's `stop` manually afterwards prints

```
nested marker invalid/dead; sockets orphan or absent — restoring idle
failed to start idle gamescope; retaining restore-idle claim
```

so the failing step looks like the idle-gamescope restore. I cannot separate "the helper has a bug" from "my hand-staged copy behaves differently from the packaged one". You have the packaged script and can settle it immediately; treat the residue as reproduced and the rc=1 attribution as unconfirmed.

## New: nested startup is slow, and much slower on reconnect

Reproducible across both runs — first session fast, every later session in the same process 3-5x slower:

| | first session | reconnect |
|---|---|---|
| run 1 | — | **69 s** |
| run 2 | **15 s** | **53 s** |

```
08:40:42.556  session_manager: waiting up to 120s for nested gamescope startup
08:41:35.573  session_manager: nested WSI session ready
```

One reconnect in run 1 exceeded the client's patience and VoidLink timed out at "connecting…" while the host was still inside that 120s window; the host itself did eventually come ready, so this is slowness, not a hang. Worth a look given the 120s allowance is doing real work here.

## New: the Steam guard leaves the desktop Steam window over the game

Hit 3 of 4 game launches across the two runs, including with host desktop Steam fully closed before connecting:

```
steam_input_guard: closed Big Picture after Steam started a tracked game     (3x)
steam_input_guard: completed [steam steam://close/bigpicture]                (3x)
```

Closing Big Picture when the tracked game starts leaves Steam in desktop mode **inside the nested session**, and its window surfaces above the game. The stream shows the Steam desktop window on top of Sonic/Quake and the game cannot be focused. It is not deterministic on a first launch: run 2's first Sonic launch was clean and is where the latency figures below come from. It then hit when a second game (Quake) was launched in that same session, hit again after reconnecting and launching Sonic, and had already hit on run 1's first launch.

This is only unrecoverable because client input does not reach the private session (below), so there is no way to dismiss the window from the client.

## #491 works, but the input leak is separate

`polaris-gamescope-session` on `PATH` does move the host off the attach path:

```
session_manager: nested WSI session ready (gamescope owned by polaris-gamescope-session)
session_event: game_launching Launching Steam Big Picture (nested WSI)
```

But the client's pointer still drives the **host desktop**, not the streamed session — so #490 was not the cause of that symptom, only of the attach fallback. The cause looks like a separate silent no-op: `client_keyboard_mouse_seat_isolation` defaults to false, and even when enabled the marker never lands, because `client_input_phys_marker()` is passed to inputtino as `device_phys` and `third-party/inputtino/src/uinput/mouse.cpp` (and `keyboard.cpp`) never call `libevdev_set_phys` — `grep -rn libevdev_set_phys third-party/inputtino/src include` returns nothing, and the only consumer of `device_phys` anywhere is the uhid PS5 path. So the udev rule at `60-polaris.rules:33`, which matches `ATTRS{phys}=="polaris/client-input-seat-isolated/*"`, can never fire. Observed directly: with both isolation options enabled, the virtual pad got `ID_SEAT=seat-polaris` (its rules match on device *name*), while both `Polaris Mouse passthrough` devices had empty `phys` and stayed on `seat0`. Separately, enabling isolation without the `input` group makes Polaris warn it cannot open the devices it just created. Happy to file this as its own issue.

## Latency, with a caveat that matters

First clean in-game 4K numbers on this host:

| 4K, Sonic Racing CrossWorlds, in-game | host processing latency |
|---|---|
| `gamescope_stream` + DMA-BUF (this opt-in) | **17-20+ ms**, higher while the game loads |
| cage/headless + SHM (`headless_mode`, `linux_capture_profile = gpu_native`) | **>30 ms** |
| Sunshine, same game and resolution, Vulkan Video + real connector | **4.7 ms** |

**These are not a clean transport comparison.** The two Polaris rows are different routes, not the same route with the transport swapped, so part of the difference may be the route rather than zero-copy. Still no `capture_telemetry` line in any run, so I have no p50/p99 for the capture stage itself. The figures are client-reported averages observed during play, and the DMA-BUF row still rises measurably during load, so contention has not gone away — it is just at a lower level than SHM.

papi-ux:
I checked this against the exact Fedora RPM retained from the v1.3.11 candidate workflow, not a source-tree staging:

- workflow run: `32211504490`
- package: `polaris-1.3.11.3b147ba-1.x86_64`
- RPM SHA-256: `b163ba318c2fb7cdcd4fdf5f0d9fc0ac386738595fe661c1fd52e581564e1cfb`

I extracted the RPM without installing it or touching the running Polaris service. Its installed `polaris-gamescope-session` and `polaris-gamescope-runtime-lib.sh` bodies match current source apart from the packaged `/usr/bin/bash` shebang.

Package-level results:

- the packaged session helper passes the full `test_gamescope_session_stop_shell.sh` stop / restore-idle state-machine contract
- the packaged session and runtime helper pass the full `test_gamescope_runtime_shell.sh` ownership and display-routing contract
- a separate isolated runtime fixture using the packaged library reclaimed all four orphan endpoints: `gamescope-0`, `gamescope-1`, `gamescope-0-ei`, and `gamescope-1-ei`
- all four Wayland `.lock` files remained, deliberately
- no `.polaris-pin.*` files remained
- a live Unix socket with a real `/proc/net/unix` row was refused and preserved, as required by the fail-closed ownership rule

So the lock files in your report are expected reusable Wayland lock state, not socket residue. The limiter files are outside #493's stated cleanup scope. The exact packaged cleanup code does remove the `-ei` endpoint when it is provably orphaned.

This does not erase your observed `gamescope-0-ei` residue. It narrows it: the package itself and its isolated stop/recovery contract pass, while your hand-staged helper could not start the installed idle service during restore. That makes the `rc=1` consistent with the staging environment, but it does not prove that attribution on your host.

The remaining decisive check is one teardown using the actual candidate RPM, with the installed user units and helper on PATH. Please record endpoint type plus any `/proc/net/unix` row before deleting anything. I am keeping the physical cleanup gate open until that packaged run is observed.

papi-ux:
I traced the reconnect timing through the current session state machine. The 53s and 69s intervals are not measuring the same path as the 15s first launch.

The outer "waiting up to 120s" timer starts before `polaris-gamescope-session start`. If durable session or nested claim state remains, `start` first runs credentialed `stop` recovery, and only then begins the new nested launch. In this run the previous teardown returned 1 and the manual retry explicitly retained a `restore-idle` claim after the staged environment could not start `polaris-gamescope-idle.service`. The next connection therefore includes that recovery attempt inside the reported startup interval.

That explains why this evidence does not yet support lowering the 120s limit or changing a readiness timeout. It also does not prove the packaged reconnect is fast. The exact RPM run is still the deciding case because it installs the helper, runtime library, idle unit, and portal units as one contract.

For that run, please retain the journal lines, with timestamps, from the outer "waiting up to 120s" line through "nested WSI session ready", especially any of:

- `complete prior session recovery before new launch`
- `tearing down marked nested WSI gamescope`
- `idle gamescope restored after nested stop`
- `nested geometry`
- `portal-gamescope rebound to nested gamescope-0`
- `nested gamescope ready`

Also record whether `polaris-gamescope-session-state`, `polaris-gamescope-session-id`, and `polaris-gamescope-wsi-nested` remain after disconnect. If the candidate RPM clears those and a clean reconnect still takes 50s or more, that is a distinct packaged startup bug and the timestamped phase will identify the bounded fix.

papi-ux:
Release decision, 2026-08-19: include this containment in 1.3.11. The maintainer accepts the remaining uncertainty that the affected RX 7800 XT has not physically tested this exact candidate. This does not claim that the underlying DMA-BUF stall is fixed. It ships the safe-default VAAPI SHM containment, while exact POLARIS_PORTAL_DMABUF=1 remains an operator opt-in for hosts that have proved the route.

CarlosSouza:
Your staging attribution is correct, and I can now name the concrete cause. It is
not a quirk of how I staged the helper — the distro package on this host predates
the packaging fix, so the gamescope contract was never installed at all.

## The host package ships none of the gamescope contract

This is Arch (CachyOS), package `polaris 1.3.10-1`, 101 files. None of them are
the session helper, the runtime library, or the idle unit:

```
$ pacman -Q polaris
polaris 1.3.10-1

$ pacman -Ql polaris | grep -cE 'gamescope|idle\.service'
0

$ ls /usr/lib/systemd/user/polaris*
/usr/lib/systemd/user/polaris.service        # the only unit installed
```

The reason is a release-ordering gap:

| | `polaris-gamescope-session` / `-runtime-lib.sh` in `packaging/linux/Arch/PKGBUILD` |
|---|---|
| tag `v1.3.10` — published 2026-08-17T04:53:09Z, **installed here** | 0 occurrences |
| `main` today | 4 occurrences (incl. the `test -x` guards) |

**#491 — "Fix Linux packages to ship Gamescope session helper" — merged
2026-08-19T01:36:32Z, two days after v1.3.10 shipped.** So every run I have
reported to you was executed against a package built before the helper was ever
packaged. That is the real content of "hand-staged": there was nothing installed
to stage against.

## Why the restore could not start the idle service

`polaris-gamescope-idle.service` is not merely absent from the package — it is
also runtime-masked on this host:

```
$ systemctl --user show polaris-gamescope-idle.service -p LoadState -p UnitFileState -p FragmentPath
LoadState=masked
UnitFileState=masked-runtime
FragmentPath=/run/user/1000/systemd/user.control/polaris-gamescope-idle.service

$ ls -la --time-style=full-iso /run/user/1000/systemd/user.control/
lrwxrwxrwx 1 <user> <user> 9 2026-08-19 08:43:27 polaris-gamescope-idle.service -> /dev/null
```

A `--runtime` mask pointing at `/dev/null`, created at the moment of the last
teardown, for a unit the package never installed. No restore path could have
started it. The `rc=1` was structurally guaranteed here, independent of #493.

## Durable state after disconnect — the check you asked for

You asked whether `polaris-gamescope-session-state`, `polaris-gamescope-session-id`
and `polaris-gamescope-wsi-nested` remain after disconnect. Captured from the
still-untouched host; the disconnect was at ~08:43 and this is intact **seven
hours later**, across no intervening Polaris run:

```
/run/user/1000/polaris-gamescope-session-state   PRESENT   "923FB184… nested"
/run/user/1000/polaris-gamescope-wsi-nested      PRESENT   "nested"
/run/user/1000/polaris-gamescope-session-id      ABSENT
/run/user/1000/polaris-gamescope.pid             PRESENT   "381519 6686028 nested /usr/bin/gamescope"
/run/user/1000/polaris-gamescope.env             PRESENT   (WAYLAND_DISPLAY=gamescope-0, ROLE=nested)
/run/user/1000/polaris-gamescope-force           PRESENT   "0"
/run/user/1000/polaris-gamescope-audio-sink      PRESENT   "sink-sunshine-stereo"
```

PID 381519 is dead — `/proc/381519` does not exist — so the pid file is stale.
No gamescope process is running and **no `gamescope-*` socket remains** in
`/run/user/1000`, so I have no `/proc/net/unix` row to report for this state:
the endpoint residue is gone while the claim files persist.

This is exactly the durable-claim precondition you described for the reconnect
timing. It survived a disconnect, hours of idle, and unrelated workloads. On the
next `start` this host would enter credentialed `stop` recovery before launching,
which matches your reading of the 53s/69s intervals — and, given the masked and
absent idle unit, that recovery could only fail again.

## On the packaged run

I cannot execute your decisive check as specified. The candidate artifact is a
Fedora RPM and this host is Arch; installing it is not a sane operation here, and
extracting it would reproduce the same hand-staged condition you have already,
correctly, declined to accept.

What I can offer is the equivalent from your own tree: build
`packaging/linux/Arch/PKGBUILD` at the candidate commit and install it with
`pacman`, giving installed user units and both helpers on `PATH` as one package
transaction — the contract you described, via the official Arch packaging rather
than the RPM. It is not the same artifact and I will not present it as one; the
binary and the packaging path both differ. Your call whether that closes the gate
or only narrows it.

If you prefer, I can also first clear the stale claim files above and capture a
clean-slate teardown on the current package, to separate "residue from a
pre-#491 package" from "residue #493 should have cleaned". I have deliberately
left every file in place pending your answer, per your request to record before
deleting.

papi-ux:
Thanks, this gives us the exact retained-state precondition we need. Please leave that runtime state untouched. Do not clear the stale claims or runtime mask before the next run.

The Polaris v1.3.11 candidate is now frozen at:

`d49a0fc8c41390d2ce65a0e0125f17b6d089f136`

Please build and install `packaging/linux/Arch/PKGBUILD` from that exact commit, using the normal Arch package transaction and `pacman`. This is the package-native test we want. The CI Arch artifacts are packaging inputs and metadata, not a portable ready-to-install package: their source path is runner-local and their commit identity is the synthetic merge commit, so please do not use those artifacts for this check.

Before launching, please record:

- installed package identity and the package file list for the session helper, runtime library, and idle service
- the idle unit's mask/load state
- the retained claim files and stale PID state, still unchanged

Then capture timestamped phases for:

1. first launch after the in-place package upgrade
2. disconnect
3. reconnect without manually clearing the retained claims

For each phase, please include the relevant helper/session logs, idle-unit state, claim-file transitions, and whether any coredump or forced kill occurred.

That will test package contents, unit recovery, stale-claim cleanup, and reconnect timing as one contract. A clean-slate run first would remove the most valuable precondition, so please preserve it.

This result can clear or narrow the Arch package path for the candidate. It does not substitute for Fedora RPM evidence and does not establish that the underlying GPU DMA-BUF issue is fixed.

papi-ux:
Small but important hold before you build: please preserve the retained runtime state exactly as it is, but do not build `d49a0fc8c41390d2ce65a0e0125f17b6d089f136` yet.

We are adding one release-publication contract so the GitHub release created beside the assets carries the curated v1.3.11 notes instead of generic generated text. That will advance the candidate commit even though it does not change the runtime fix. I will post the replacement exact SHA after its checks and review are complete.

Nothing else about the requested Arch procedure changes.

papi-ux:
The release-publication work is complete, and the replacement Polaris v1.3.11 candidate is now frozen at:

`2c41a569fc6a67fe42b9415aabbecb1f4cb126ae`

The hold from the prior comment is lifted. Please proceed with the package-native Arch procedure from https://github.com/papi-ux/polaris/pull/481#issuecomment-5348518263, using this replacement exact commit.

Please continue to preserve the retained runtime state exactly as recorded. Build and install `packaging/linux/Arch/PKGBUILD` from this commit through the normal Arch package transaction and `pacman`; do not substitute the CI artifacts. Record the same preflight package identity, helper/runtime/unit contents, idle-unit state, retained claims, stale PID, timestamped first-launch/disconnect/reconnect phases, claim transitions, teardown outcome, coredumps, and any forced kill.

All applicable exact-head checks are green at this SHA, including the Arch build, full web suite, sanitizer, Ubuntu, SteamOS, both Fedora lanes, CodeQL, and public hygiene. Independent review found no remaining material issue in the final candidate.

Nothing else about the requested procedure changes.

CarlosSouza:
Package-native Arch run completed on the replacement candidate. Summary first: the
packaged path reproduces the `rc=1` teardown, and the reconnect does not merely
run slow — it fails, with two `gamescope` SIGABRT coredumps.

Host: CachyOS, Linux 7.1.8-1-cachyos, AMD Radeon RX 9070 XT (radeonsi, gfx1201),
Mesa 26.1.6-arch3.1, KDE Plasma 6.7 Wayland, GCC 16.2.1, boost-libs 1.91.0.
`linux_stream_mode = gamescope_stream`, `encoder = vaapi`, `capture` unset.

## Provenance and disclosed deviations

Built `packaging/linux/Arch/PKGBUILD` at `2c41a569fc6a67fe42b9415aabbecb1f4cb126ae`
through `makepkg` + `pacman -U`. No CI artifacts used. The running binary confirms
the commit itself:

```
Info: Polaris version: 1.3.11 commit: 2c41a569fc6a67fe42b9415aabbecb1f4cb126ae
```

Two deviations were required, both matching what @FabianZettl reported on #480 for
the same compiler, and both unrelated to the fix under test:

- `BUILD_WERROR=OFF`
- the GCC 15 `_FORTIFY_SOURCE` filter extended to GCC 16 (`=~ ^(15|16)$`)

The template placeholders were filled locally (`pkgver=1.3.11`, clone URL, commit).
boost here is still 1.91.0, so the drift @FabianZettl hit did not apply.

## Preflight: package contents and unit state

`#491` is satisfied — both helpers are package-owned and on `PATH`:

```
/usr/bin/polaris-gamescope-session          36716 bytes
/usr/bin/polaris-gamescope-runtime-lib.sh   27018 bytes
```

Both are **byte-identical** to the helpers I hand-staged in the earlier report
(`sha256` prefixes `3dff8955…` and `95d31f14…`). The staged content was never the
variable; only installation and the surrounding contract were.

One contract difference from your description of the RPM: on Arch the package ships
**only** `polaris.service` under `systemd/user`. There is no packaged
`polaris-gamescope-idle.service` and no packaged portal units —
`cmake/packaging/linux.cmake` installs the two helpers, the udev rule and the
modules-load conf, and `polaris.service`. The idle unit is written to
`~/.config/systemd/user/` by the `--setup-host` flow instead. Throughout this run:

```
polaris-gamescope-idle.service   LoadState=not-found   UnitFileState=(empty)
```

So "unit recovery" could not be exercised as one packaged contract here. If the RPM
does install that unit, the two distro paths differ in scope.

## Why my earlier report showed nested state at all

Your staging attribution was right, and here is the mechanism. On the stock
pre-#491 package the nested path is **unreachable**, stated by the code itself:

```
Warning: session_manager: cannot nest Steam Big Picture because
         polaris-gamescope-session is not on PATH; using attach
```

1.3.10 therefore only ever reaches `role=runtime`; a full session left
`polaris-gamescope.pid = "66420 3965209 runtime /usr/bin/gamescope"` with
`session-state` and `wsi-nested` absent. The `nested` claims in my earlier report
existed only because the staged helpers made that path reachable. Staging was not
gratuitous — it was the only way to reach the code path under discussion.

Note also that `pacman -U` does **not** swap the running host. After the upgrade the
service was still executing the old image:

```
/proc/65072/exe -> /usr/bin/polaris-1.3.10 (deleted)
```

A `systemctl --user restart polaris.service` was required for 1.3.11 to run. That
restart killed the leftover gamescope and removed its sockets but **kept** the claim
files, producing precisely the stale-claim precondition you asked for: `.pid`
pointing at dead PID 66420, plus `.env`, `-force`, `.lock`.

## Phase 1 — first launch after the in-place upgrade: PASS, 15.0s

```
10:18:19.906  session_manager: nested WSI SDR path — polaris-gamescope-session start
              (Steam Big Picture primary child; not attach/X11)
10:18:20.047  Executing Do Cmd: [/usr/bin/polaris-gamescope-session start]
10:18:20.049  session_manager: waiting up to 120s for nested gamescope startup
10:18:35.058  session_event: game_launching  Launching Steam Big Picture (nested WSI)
10:18:35.058  session_manager: nested WSI session ready (gamescope owned by
              polaris-gamescope-session)
```

15.0s from the outer timer to ready, using the **packaged** helper. The stale claims
did not trigger any logged recovery step — no `complete prior session recovery
before new launch` line appears anywhere in this run. A full Sonic race was played.

Two observations from this phase that are not teardown-related:

**(a) The #481 containment is active on this host and costs it the DMA-BUF path.**

```
10:18:39.583  portal: vaapi_pipewire_dmabuf_disabled_for_stability source=local_graph;
              offering SHM by default; set POLARIS_PORTAL_DMABUF=1 only on a host
              where this path is known to work
10:18:39.606  dmabuf_negotiate may_use=false force_cpu=false has_modifier=false
              offered=false negotiated=false
10:18:39.694  Warning: portal: capture_transport=shm frame_residency=cpu frame_format=bgra8
```

For contrast, the same host on 1.3.10 minutes earlier negotiated
`capture_transport=dmabuf frame_residency=gpu`. This is the documented trade-off, not
a defect — recorded because this host is one that proved the route.

**(b) Nested geometry is a hardcoded 4K120 default, independent of the client.** The
client streamed 1080p60; the nested compositor was created at 4K120 and capture ran
at 4K:

```
gamescope --backend headless --steam --xwayland-count 2 -W 3840 -H 2160 -r 120 \
          -w 3840 -h 2160 -- steam -gamepadui
portal: Initializing capture 1920x1080 client_dynamic_range=0
portal: Capture ready — 3840x2160 env=3840x2160
```

The source is `/usr/bin/polaris-gamescope-session` lines 21-23:

```bash
gs_width="${POLARIS_HDR_WIDTH:-3840}"
gs_height="${POLARIS_HDR_HEIGHT:-2160}"
gs_refresh="${POLARIS_HDR_REFRESH:-120}"
```

Those three variables do not appear in the `polaris` binary at all (checked with
`strings`) and are not set in `apps.json`, so nothing overrides them — every nested
WSI session gets a 4K120 compositor regardless of what the client negotiated.

I first suspected a stale mode left by an earlier 4K120 session from a different
host on the TV, and ruled it out: the dummy output carrying that mode
(`HDMI-A-1`, `3840x2160@120.00`) is **disabled**, the only enabled output is
`DP-3` at `2560x1440@143.97`, and Polaris itself logged `Resolution: 2560x1440` at
startup. Neither the active desktop nor the client's request is 4K120.

Combined with (a) that is a 4K CPU-side readback feeding a 1080p stream — roughly
four times the pixels the client asked for, through the copy path. Measured host
processing was 17-18ms for most of the race, with audible audio choppiness. No
PipeWire xrun/underrun is logged, so I am reporting the audio as an observation
without a host-side signature. Happy to file (b) separately — it looks independent
of the teardown failures below, and a 1080p60 client paying 4K120 capture cost by
default seems worth its own issue.

## Phase 2 — disconnect: `rc=1` reproduces on the packaged path, plus a coredump

```
10:23:42.879  Skipping Steam -shutdown undo; polaris-gamescope-session stop owns
              nested Steam cleanup
10:23:42.880  Executing Undo Cmd: [/usr/bin/polaris-gamescope-session stop]
10:23:42.885  session_manager: waiting up to 30s for nested gamescope teardown
10:23:44.885  Warning: Return code [1]
```

**This is `/usr/bin/polaris-gamescope-session` from the candidate package, invoked by
the candidate binary, with the packaged unit.** The `rc=1` is not a property of my
staging environment.

`gamescope` aborted during this teardown:

```
PID: 77687 (gamescope-wl)   TID: 77766 (gamescope-xwm)
Signal: 6 (ABRT) si_code: SI_TKILL
User Unit: polaris.service
Stack: pthread_kill → raise → abort → gamescope + 0xba2f5
```

The nested WSI log gives the mechanism:

```
[gamescopereaper] reaper: Parent of gamescopereaper was killed. Killing children.
[gamescope] wlserver: [xwayland/server.c:217] Restarting Xwayland
[gamescope] Error: xwm: X11 I/O error! This is fatal. Aborting...
```

The reaper's parent is killed, the reaper kills children, Xwayland dies, and the XWM
thread aborts on the resulting X11 I/O error — consistent with the `gamescope-xwm`
thread in the coredump.

Residue immediately after this disconnect: `gamescope-0` and `gamescope-0-ei` still
present as files but with **no `/proc/net/unix` rows** — orphaned endpoints, not live
sockets — alongside `session-state = "437AF03E… nested"`, `wsi-nested = nested`, and
`.pid = "77687 4011836 nested /usr/bin/gamescope"` with 77687 dead.

## Phase 3 — reconnect with the retained claims: FAIL

Nothing was cleared between phases, per your instruction.

```
10:27:45.244  Executing Do Cmd: [/usr/bin/polaris-gamescope-session start]
10:27:45.246  session_manager: waiting up to 120s for nested gamescope startup
              ← log then goes silent for ~111s
10:29:36      gamescope PID 151371 SIGABRT (second coredump)
10:29:37.255  Executing Undo Cmd: [/usr/bin/polaris-gamescope-session stop]
10:29:37.264  session_manager: waiting up to 30s for nested gamescope teardown
10:30:07.267  Warning: Return code [383]
10:30:07.270  session_event: session_ending [tearing_down] Cleaning up
10:30:07.272  session_manager: Session state restored
```

`nested WSI session ready` never appears. The client timed out. So the answer to
whether the packaged reconnect is fast is neither "fast" nor "50s or more" — it does
not complete.

Two further observations during this phase:

- **Concurrent stop invocations.** Two `polaris-gamescope-session stop` processes were
  live simultaneously (PIDs 166582 and 193001) while the hang was in progress.
- The second teardown returned **383**, not 1. `383 & 0xFF == 127`, which would be a
  "command not found" wait status; I did not chase that further.

Total for the run: **2 `gamescope` SIGABRT coredumps**, both under
`polaris.service`, at 10:23:44 (PID 77687) and 10:29:36 (PID 151371). No forced kill
was issued by me at any point in phases 1-3.

## Final residue

```
gamescope-0, gamescope-0-ei                  REMOVED          ← #493 endpoint cleanup worked
gamescope-0.lock, gamescope-0-ei.lock        remain           ← expected reusable lock state
gamescope-limiter-XXuvMOrl (10:18)           remains
gamescope-limiter-XXT7BZal (10:28)           remains          ← 2 accumulated, never reclaimed
polaris-gamescope-session-state              remains          "C9C36857… nested"
polaris-gamescope-wsi-nested                 remains          "nested"
polaris-gamescope.pid                        remains          "151371 … nested", 151371 dead
polaris-gamescope.lock (10:10)               remains          ← survived from the 1.3.10 era
polaris-gamescope-session-id                 never created in any phase
```

So #493's endpoint reclamation does work on the packaged path — both sockets were
removed and no `/proc/net/unix` row was left behind. What persists is the **claim
set**, which is what feeds the next connection into recovery, and the limiter files,
which accumulate one per session.

## What this does and does not establish

It establishes that on this Arch host, with the candidate package and no staging:
the nested path works and is fast on first launch; the teardown returns `rc=1` and
aborts `gamescope`; and a reconnect over the retained claims fails outright with a
second abort.

It does not establish anything about the Fedora RPM, does not test unit recovery
(no packaged idle unit on Arch), and does not speak to the underlying GPU DMA-BUF
question — this host takes the SHM default now, by design.

All artifacts retained: per-phase state captures, both coredump reports, the nested
WSI logs, and the full host logs for the 1.3.10 precondition and the 1.3.11 phases.
Happy to attach any of them, run a patch, or re-run with `POLARIS_PORTAL_DMABUF=1`
to separate the transport variable from the teardown failures.

papi-ux:
Thank you for the package-native run. This separates three outcomes clearly:

- the VAAPI containment from this PR is active and the SHM path delivers media
- the endpoint cleanup from #493 works
- nested teardown and reconnect still fail as a separate lifecycle defect

I am tracking the lifecycle failure in #507. The hardcoded nested geometry is tracked independently in #505 so neither change obscures the other.

The lifecycle fix will serialize ownership transitions, attempt exact-session Steam shutdown before compositor-group termination, retain the existing identity checks, and add repeated-stop and recovery coverage. Package-native affected-host validation with no coredump remains required before that issue can close.

CarlosSouza:
Follow-up on the `POLARIS_PORTAL_DMABUF=1` opt-in, measured on v1.3.13 on the same
RX 9070 XT host. **The result argues for your safe default, not against it.**

The opt-in works exactly as designed:

```
portal: vaapi_pipewire_dmabuf_explicitly_enabled source=local_graph; operator opted
  into an unvalidated path with no automatic stall fallback
portal: dmabuf_negotiate may_use=true force_cpu=false has_modifier=true modifier=0
  spa_format=8 formats=4 offered=true negotiated=true flags=4
portal: capture_transport=dmabuf frame_residency=gpu frame_format=bgra8
```

GPU-resident DMA-BUF, negotiated cleanly, no stall across a full race at 4K60,
80 Mbps, `hevc_vaapi`.

**But it buys essentially nothing measurable.** Client-reported host processing:

| transport | menus | during a race |
|---|---|---|
| SHM (default) | ~12ms | 20-30ms |
| DMA-BUF (opt-in) | 11.8ms | 20-30ms |

I expected a clear win and was wrong. The tell is that the cost is load-dependent: a
capture-copy cost would be constant, since 4K is the same ~31.6 MiB per frame whether
the scene moves or not. A cost that only appears under motion is the encoder. At 4K60
the `hevc_vaapi` encode dominates and the capture transport is noise next to it.

Two implications, both in favour of what you shipped:

1. On a host that demonstrably supports the DMA-BUF route, the SHM containment costs
   no measurable latency at 4K60. The safe default is not trading performance for
   safety here — at least not at this resolution on this GPU.
2. The opt-in's own warning is the accurate description of the deal: an unvalidated
   path with no automatic stall fallback, for no measurable gain. I have reverted it
   on this host and would not recommend it to another RDNA4 operator on these numbers.

Where the real gap is, for context rather than as a request: Sunshine on this same
machine and resolution reports 4.7ms, using FFmpeg **Vulkan Video** encode rather than
VA-API. We measured VA-API carrying roughly 5.8ms of fixed per-frame cost here
earlier. That difference is the encoder backend, not capture — which is consistent
with the table above.

Unrelated but worth one line since it recurs in every session on this host:
`av1_vaapi` measured above 20ms in menus where `hevc_vaapi` sat at ~12ms, and logs
`Multiple slices were requested but this codec does not support controlling slices`.

PR #512, opened 2026-08-20
feat(doctor): offer a one-click fix for the Steam Input conflict

papi-ux:
## Summary

- Give the `steam_input_conflict` finding from #504 the guarded one-click action the rest of Doctor already has. The finding shipped read-only: it named the setting and left the user to go find it.
- Action id `disable_steam_input_xbox`, kind `host_setting`, label **"Close Steam and fix"**. It requires confirmation and owner, is never destructive, and supports undo.
- **Applying it closes desktop Steam first, which is why it confirms.** Steam holds its profile in memory and rewrites `localconfig.vdf` on exit, so an edit made underneath a running client is reverted the moment that client closes. It reuses `proc::request_desktop_steam_shutdown_for_private_stream()`, which already waits for Steam to go quiescent and reports whether it did — so a refusal to close is a refusal to apply, never a change that silently disappears later.
- **Verification reports three outcomes, not two.** Clearing the host-wide opt-in while per-game Force On overrides remain is a real partial result: those overrides outrank the setting this action owns and must be changed per game. Saying so beats claiming a success the controller will not show.
- The write is surgical and atomic: `set_steam_input_xbox_support()` replaces exactly one value and preserves the rest byte for byte, and the payload is swapped in by `rename`. A half-written `localconfig.vdf` is a corrupted Steam profile — a much worse outcome than the setting we came to change.
- Undo restores the previous value on every profile the run touched.

## Two placements worth reviewing

1. **The action runs ahead of the `!stats.streaming` gate** that every other action sits behind. Its purpose is to edit a profile Steam is not holding open, so it applies exactly when nothing is streaming — the inverse of the live bitrate steps. `verify` therefore asks which run is in flight before that gate decides a stream is required.
2. **Verification bypasses the 30-second snapshot cache** via an uncached re-inspection. Verify runs 2s after the write, well inside that cache, so the cached snapshot would report the state this action just changed and call its own success a failure.

## Re-pinned guard

`StreamStatsDoctorTests.WarnsWhenSteamInputConflictsWithStrictIsolation` pinned `safe_recovery_action.id == "none"`, asserting the finding was action-free. That intent is deliberately superseded here, so it is re-pinned in its new form rather than deleted: the action must stay reversible, never destructive, never applied without confirmation, owner-gated, and routed at `/api/doctor/action`.

## Exact candidate

- `Commit:` `667f99db6f1a240b47a41b769d6060a0157811c3`
- `Tree:` `a5805a7c11eb7829a144dc5c31415a5995fb92c2`
- `Parent:` `7d75db9e`
- `Base:` `7d75db9e` (polaris/master, carries #504 and #510)

## Verification

- [x] `ctest` full suite — 17/17 passed, 0 failed
- [x] `test_polaris --gtest_filter='SteamInputTests.*:SteamPlaytimeTests.*'` — 13/13, including three new cases for the writer
- [x] `test_polaris_stream --gtest_filter='StreamStatsDoctorTests.*'` — 16/16
- [x] The new writer tests cover: one value changes and the payload length is unchanged; undo round-trips to the original byte for byte; no write when the value already holds the request or the key is absent; and a key written without a value does **not** borrow the next line's value and rewrite the wrong field.

## Not covered

No unit test drives `doctor_actions::execute` for this action end to end. Doing so would require a real Steam client to close and real profiles to rewrite; this matches the existing pattern in that file, where the route handlers and Steam lifecycle paths are untested glue over tested pure logic. The pure parts — the payload transform and the aggregate inspection — are covered directly. The action has **not** been exercised on the live host, because the host it was written for has already had its conflict fixed by hand and no longer reproduces the finding.

### DIFF
--- a/src/doctor_actions.cpp
+++ b/src/doctor_actions.cpp
@@ -8,13 +8,22 @@
 #include <algorithm>
 #include <chrono>
 #include <cmath>
+#include <filesystem>
+#include <fstream>
 #include <mutex>
+#include <sstream>
 #include <string>
+#include <vector>
 
 #include <nlohmann/json.hpp>
 
 #include "adaptive_bitrate.h"
+#include "game_library_scanner.h"
 #include "logging.h"
+
+#ifdef __linux__
+  #include "process.h"
+#endif
 
 using namespace std::literals;
 
@@ -23,13 +32,21 @@ namespace doctor_actions {
     enum class action_kind_e {
       none,
       lower_bitrate,
-      restore_quality
+      restore_quality,
+      disable_steam_input_xbox
+    };
+
+    /** One profile this run rewrote, and the value it held beforehand. */
+    struct steam_profile_edit_t {
+      std::filesystem::path path;
+      bool previously_enabled = true;
     };
 
     struct action_run_t {
       bool active = false;
       action_kind_e kind = action_kind_e::none;
       std::string run_id;
+      std::vector<steam_profile_edit_t> steam_edits;
       bool previous_adaptive_enabled = false;
       int previous_base_bitrate_kbps = 0;
       int previous_live_bitrate_kbps = 0;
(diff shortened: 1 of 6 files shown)


papi-ux:
Pushed `87ed80b4`: the confirmation dialog now names its own consequence.

The web dialog renders `label` followed by `rollback`, and neither said that closing Steam ends the game running through it. Because the finding is only live while a Steam game is streaming, the moment this action is reachable is exactly the moment pressing it ends that session. A confirmation that hides its own cost is worse than no confirmation, so the rollback text now leads with it.

Updated `Commit:` `87ed80b4de1afb9b27a8b7339aac79189315a5a5` / `Tree:` `783646925cc60e33ede7f8eaa2135ed632068d47`. StreamStatsDoctorTests still 16/16.

Noting the underlying tension for whoever reviews: the action has to be usable during a stream because the button lives on a finding that only exists during one, yet the edit it makes is only safe with Steam closed. Warning is the honest resolution for this PR. The cleaner long-term fix is to keep the Steam Input snapshot alive outside sessions so the finding — and its button — survive into idle, where closing Steam costs nothing.

PR #545, opened 2026-08-27
fix(linux): close virtual display ownership gaps

### LINKED ISSUE #532: [Bug]: Mirror desktop mode, with virtual display as application creates a new monitor display but I have no way to access it from host or client
### Describe the bug

Creating a mirror desktop connection will create a second display but I cannot access it from either host or client.

### Steps to reproduce

Polaris setup with mirror desktop runtime, client select virtual desktop as the application

### Install method

Source build

### Host OS

Omarchy 4.0.0-1 - Hyprland 0.56.2 (Wayland)

### GPU and driver

AMD 9070XT

### Client and device

Moonlight on iPad

### Runtime path

None

### Expected behavior

Virtual display should show the second monitor on client device instead of the host display. Runtime is mirror desktop though so theoretically it's correct, but I can't access the new virtual desktop anyhow.

### Relevant log output

```Shell

```

### Screenshots or recordings

Image

[Log.md](https://github.com/user-attachments/files/31344231/Log.md)

### Additional context

_No response_

papi-ux:
## Summary

- make bundled and narrowly matched legacy Desktop launches explicitly mirror the desktop even when paired settings prefer a virtual display
- bind portal, KWin, and wlroots capture plus reinitialization to the exact launch generation, output, private Wayland socket, and runtime instance
- make Linux virtual-display creation and teardown transactional: durable pre-mutation intent, tri-state readback, stable Hyprland absence, exact KScreen topology restoration, and fail-closed stale cleanup

## Exact scope

- base: 6738b2e60a213aceb6434c77e164bcbda48b6b2d
- head: dd8aca8a51411100da44d0dfcad79b4dfdd3d5dd

## Validation

- exact-head Linux build completed
- full configured native suite passed 18/18
- focused Desktop migration, virtual-display, capture-generation, source-safety, stream, platform, config, and video coverage passed
- independent exact-head review: CLEAN

Fixes #532

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,7 @@
 # Changelog
 
 - Fixed Gamescope session recovery on distro-package hosts by restoring their empty compositor baseline without requiring Nix-only idle/portal units, and made failed prior-session recovery retain its exact claim instead of forcing unsafe socket cleanup.
+- Fixed Linux host-virtual-display capture so each launch is bound to the exact output it created, while Desktop mirror launches stay on the desktop path even when an app or optimizer prefers a virtual display. Capture now fails closed instead of silently selecting another monitor.
 
 This file tracks the public Polaris release line.
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,7 +1,7 @@
 # Changelog
 
 - Fixed Gamescope session recovery on distro-package hosts by restoring their empty compositor baseline without requiring Nix-only idle/portal units, and made failed prior-session recovery retain its exact claim instead of forcing unsafe socket cleanup.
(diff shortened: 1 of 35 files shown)


FabianZettl:
Validated this against a real, unmodified Moonlight client (not just Nova) — built from current `master` and confirmed both a `desktop-mirror: true` apps.json app and the paired-settings virtual-display path now correctly force `path=desktop_display`, `virtual_display=false` on the physical output, even though the client itself has no way to request that mode explicitly.

Before this PR (still on tagged `v1.3.13`), those same apps.json entries always fell back to a virtual/headless display when launched from standard Moonlight — `apply_app_display_semantics` wasn't in that release yet. Works as intended once built past this commit. Thanks for the fix!

PR #561, opened 2026-08-31
fix(linux): keep nested Gamescope behind the teardown fence

papi-ux:
## Summary

- run Steam beneath a session-owned Gamescope primary-child keeper
- pause and revalidate only the exact marked compositor before asking exact-session Steam to exit
- keep the keeper runnable so it can publish the matching terminal marker while Gamescope is stopped
- commit the complete private process-group freeze/SIGKILL fence without any failure path that can resume Gamescope over a retired Xwayland connection
- bind the keeper to Gamescope with a parent-death signal and retain that exact wrapper as cleanup authority until every surviving private-session member is terminal
- recover Xwayland readiness after the exact private generation is fenced

## Why

Two exact candidates passed CI and then failed the required physical gate on Gamescope 3.16.25:

1. `ac84b0d0` let Steam exit naturally beneath Gamescope. Four seconds after End Session began, Gamescope produced a fresh SIGSEGV in `CVulkanDevice::~CVulkanDevice` while the exact fence was still acquiring authority.
2. `2f01943b` kept Gamescope's primary child alive after Steam exited, but still asked Steam to terminate before pausing the compositor. Five seconds after End Session began, Gamescope produced a fresh SIGABRT through `handle_io_error`, `_XIOError`, `XPending`, and `xwayland_ctx_t::Dispatch` when Xwayland disappeared before the fence took control.

Those results cover both sides of the affected upstream exit-crash family tracked in [ValveSoftware/gamescope#1526](https://github.com/ValveSoftware/gamescope/issues/1526): returning from the primary child can enter the Vulkan destructor, while removing Xwayland under a running compositor can enter the X11 I/O abort path.

This revision moves the mutation boundary. Polaris first sends SIGSTOP to the exact marker-bound Gamescope PID and verifies kernel stop state without stopping its wrapper or session Steam. It then asks only credential-bound Steam to exit and requires the runnable primary-child keeper to publish the matching session marker. Finally, it freezes and SIGKILLs the already-authorized complete private group. Once Steam teardown begins, no error path may resume the compositor.

## Exact automated verification

Exact head: `6e5c9222b03d10d006a2b37d4af61f458a9d0c41`

- local diff validation, Bash parsing, ShellCheck, C warnings-as-errors, and the cross-platform primary-child regression passed
- the real Linux fixture creates a stubborn member in a leaderless sibling process group and requires the parent-death fence to retire it by exact positive PID
[...]

PR #563, opened 2026-09-01
Fix Gamescope primary-child reaper ancestry

papi-ux:
## Summary

- model the Gamescope 3.16 process tree as Gamescope -> gamescopereaper -> primary wrapper
- validate exact executable, parent PID, start time, process group, and session identity before Steam starts
- retain the parent-death cleanup fence after Steam returns
- make the Linux fixture reproduce the upstream reaper layer and keep the leaderless-descendant teardown checks

## Why

The exact merged candidate passed CI but failed physical launch before Steam started. The helper treated the wrapper as a direct Gamescope child; Gamescope 3.16.25 actually starts the command through gamescopereaper. The false rejection made the primary command return and drove Gamescope into the shutdown path this containment was intended to avoid.

Upstream process implementation:
- https://github.com/ValveSoftware/gamescope/blob/3.16.25/src/steamcompmgr.cpp#L8296-L8316
- https://github.com/ValveSoftware/gamescope/blob/3.16.25/src/Apps/gamescopereaper.cpp#L131-L157

## Verification

- focused Linux primary-child production-shape fixture passes
- Gamescope parent death still drains Steam, an ordinary descendant, and a leaderless sibling-group member
- runtime ownership and display-routing shell suite passes
- shell syntax and diff checks pass

## Scope

This is a narrow follow-up to #561. It does not change Doctor, AI, launch presets, bitrate policy, or release publication. The failed candidate remains rolled back; physical reactivation waits for exact-head review and CI.

### DIFF
--- a/nix/modules/polaris-gamescope-runtime-lib.sh
+++ b/nix/modules/polaris-gamescope-runtime-lib.sh
@@ -85,6 +85,36 @@ polaris_headless_gamescope_pid() {
   POLARIS_GAMESCOPE_EXECUTABLE="$exe_path"
 }
 
+# Gamescope 3.16 does not exec the command after `--` as its direct child.
+# It starts gamescopereaper, which becomes a subreaper and then starts the
+# primary command. Keep that process explicit in the ownership chain: treating
+# the command wrapper as a direct Gamescope child rejects the normal upstream
+# launch shape and makes Gamescope enter its unsafe primary-child shutdown.
+polaris_gamescope_reaper_pid() {
+  local pid="$1" arg first=1 executable=""
+  local exe_path exe_name
+  exe_path="$(readlink "$(polaris_proc_root)/$pid/exe" 2>/dev/null)" || return 1
+  exe_name="${exe_path##*/}"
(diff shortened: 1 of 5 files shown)


papi-ux:
Independent exact-object review result: **PASS** at `d2c92ef10113f21ba91b1583f6037e50d661203c` over `78ccac28ba7a3cebe19d98eb45fa1f2b50967d8b`. The review verified the upstream Gamescope -> gamescopereaper -> primary-command chain, double PID/start/PPID/PGID/SID binding before Steam, retained parent-death authority after Steam exits, positive-PID leaderless cleanup, and package inclusion. No Linux containment release blocker or Doctor/AI/limiter drift was found. Non-blocking follow-up: the fake reaper could model upstream subreaper/SIGTERM forwarding more literally; current product behavior was source-verified and both active-Steam and post-Steam parent-death cleanup are exercised. Exact-head sanitizer/ownership, Nix, Arch, rolling Arch, Fedora RPM/Clang, web, CodeQL, and hygiene are green; Ubuntu and SteamOS fanout remain in progress.

Issue #571, opened 2026-09-02
Mission Control: Video/Audio page mixes read-only status with editable controls


papi-ux:
## Summary

The Video/Audio settings page in Mission Control mixes read-only status with editable controls and styles them nearly identically, so users cannot tell what they are allowed to change. Reported by u/Zerwin on Reddit against 1.4.0 (the post-rework layout), so the recent settings rework did not resolve this page:

https://www.reddit.com/r/linux_gaming/comments/1w41olv/comment/p7dbvk5/

## Specifics from the report

- "Manual stream tuning" renders as 4 boxes, none of which can be changed.
- "Display Planner - Best for this device" renders as 3 boxes with very similar styling, but there one box is selectable.
- Some boxes contain what looks like a text field but does not accept input (leftover styling from before the launch-mode expansion).
- The page reads like documentation squeezed into a settings surface; the density makes it hard to understand what each item means.
- The confusion is mostly limited to the Video/Audio page. The other settings pages read fine.

## Direction

Make the split obvious: settings look like settings, status looks like status.

- Read-only provenance/status blocks (Manual stream tuning) get a distinct non-interactive treatment. Nothing renders as a text field unless it accepts input.
- Selectable choices (Display Planner) get clear affordances for "this is a choice" and "this one is active".
- Move explanatory copy out of the page where it duplicates the docs; the launch-modes doc (https://papi-ux.com/docs/launch-modes/) already carries the deeper material, link to it instead.

## Verification

- A user can identify every element on the Video/Audio page as either changeable or informational at a glance, without clicking.
- No non-editable element renders with text-field styling.
- Follow up in the Reddit thread once the rework lands.

papi-ux:
Shipped across the incremental slices already on master (#574, #575, #576, #578, #579) and the integration PR #584.

What changed on the Video/Audio page:

- Controls, status, and reference prose are three visual grammars now and never mixed. Editable fields use one shared input style, host status uses stat tiles that are never focusable, and long-form explanations moved to papi-ux.com/docs with a one-line pointer.
- "Manual stream tuning" is gone as four look-alike boxes. Without host data the strip shows your saved settings; with a Polaris host that serves the new settings projection it shows live host status: state, live and target bitrate, smoothed round trip and loss, and the blocking reason when there is one.
- The Display Planner cards are real buttons with cursor, focus ring, and pressed state; the recommendation is labelled "Balanced" for what it is, and the chosen plan survives a reload.
[...]

PR #577, opened 2026-09-02
refactor: extract shared settings metadata builders from nvhttp

papi-ux:
## Summary

Part of #571 (settings revamp arc, backend slice 2). Mechanical extraction, no behavior change except one additive JSON key.

- New `src/settings_metadata.{h,cpp}` unit holding the builders the web UI's server will project in the next slice: the mode-selection facades, `stream_display_mode_options_json`, `build_auto_quality_policy_json` + `auto_quality_blocked_reason`, `host_virtual_display_available`, and a new `build_tuning_json` extracted verbatim from the session-status handler's inline tuning block. Bodies moved unchanged; 24 nvhttp call sites qualified.
- `mode_option_t` gains `badge`, populated from the stream_path registry descriptor, so the host's real badges (including "Recommended" on `headless_stream`) reach the mode options JSON instead of being dropped. Additive key only; Moonlight-lineage clients ignore unknown keys. The non-Linux fallback branch mirrors the registry values.
- `vDisplayDriverStatus` in the moved Windows branch is now `proc::`-qualified, matching every other consumer (the old unqualified spelling had no using-directive to resolve it).
- Two source-guard tests that pinned moved bodies by nvhttp source text are re-pointed at `settings_metadata.cpp` with their invariants intact (notably: the two-argument effective-selection overload must not read live proc session state).
- New `test_settings_metadata.cpp` (TEST_STREAM_SOURCES): blocked-reason matrix, off-state policy contract (the only reachable state while the legacy AI flag is hardcoded off - not inventing reachability), tuning JSON 14-key pin, badge propagation.
- CMake: both new files in `POLARIS_TARGET_FILES`; test registered in `tests/CMakeLists.txt`.

## Exact candidate

- `Commit:` 4ba45e5c3876c74a0de9803ee6df8b092c0e0658
- `Tree:` e2697eafc5c269d6ce573fd54e6aa7bc2877cfd9
- `Parent:` 1b0a6fd86efb1d60352ef1ae28a646771bf4cd11
- `Base:` 1b0a6fd86efb1d60352ef1ae28a646771bf4cd11

## Verification

- [x] `cmake .` + `ninja polaris test_polaris test_polaris_stream` — clean
- [x] `test_polaris_stream --gtest_filter="*SettingsMetadata*"` — 5/5
- [x] `test_polaris --gtest_filter="*DoctorResetContract*"` — 35/35 (re-anchored guard)
- [x] `ctest -R "test_polaris_stream|test_polaris_http|test_polaris_config" -j 3` — 3/3 under BUILD_FULL_TESTS=ON
- Windows/#else branches compile-checked by inspection only (no Windows builder here); CI's matrix covers them.

### DIFF
--- a/cmake/compile_definitions/common.cmake
+++ b/cmake/compile_definitions/common.cmake
@@ -195,6 +195,8 @@ set(POLARIS_TARGET_FILES
         "${CMAKE_SOURCE_DIR}/src/thread_safe.h"
         "${CMAKE_SOURCE_DIR}/src/sync.h"
         "${CMAKE_SOURCE_DIR}/src/round_robin.h"
+        "${CMAKE_SOURCE_DIR}/src/settings_metadata.cpp"
+        "${CMAKE_SOURCE_DIR}/src/settings_metadata.h"
         "${CMAKE_SOURCE_DIR}/src/stat_trackers.h"
         "${CMAKE_SOURCE_DIR}/src/stat_trackers.cpp"
         "${CMAKE_SOURCE_DIR}/src/stream_recorder.h"
--- a/src/nvhttp.cpp
+++ b/src/nvhttp.cpp
@@ -89,6 +89,7 @@
 #include "client_profiles.h"
 #include "client_support_report.h"
 #include "confighttp.h"
+#include "settings_metadata.h"
 #include "stream_stats.h"
 #include "video.h"
 #ifdef __linux__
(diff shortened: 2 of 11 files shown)


papi-ux:
Superseded: this content (plus the nova-contract manifest fix its sanitizer round surfaced) is folded into the consolidated integration/settings-revamp branch per papi's call to ship the remaining arc as one PR instead of the incremental train. Branch retained.

PR #591, opened 2026-09-03
ci: resolve the build source to a published commit on pull requests

### LINKED ISSUE #590: SteamOS 3.8 build fails when a pull request is merged while the job is still running
## What happens

`SteamOS 3.8 build` failed on #584, #586, and #588 with makepkg unable to read a
tree while creating its working copy:

```
  -> Creating working copy of polaris git repo...
fatal: unable to read tree (9dc7168134a16bbdd4b2c01ed342c2d6bc11d759)
```

## Why

The object is GitHub's synthetic merge commit for the pull request. GitHub
advertises `refs/pull/<n>/merge` while the pull request is open, so a clone can
resolve it, and it stops advertising the ref once the pull request is merged.

The SteamOS job clones its source several minutes into the run, after the chroot
is assembled and packages are installed. Merging the pull request before the job
reaches that step deletes the ref it is about to ask for.

| PR | SteamOS clone | merged | result |
| --- | --- | --- | --- |
| #584 | after merge | 23:55:37 | failed |
| #586 | after merge | 02:24:11 | failed |
| #588 | 03:45:48 | 03:42:06 | failed |
| #589 | 04:46:02 | 04:53:23 | passed |

#589 is the control: same workflow, same kind of change, merged after its clone
step, and green. Every failure was a pull request merged roughly a minute after
the job started, which is well inside the window before the clone.

The four required checks are public-hygiene, Web checks, C++ sanitizer tests, and
Arch build. SteamOS is not among them, so nothing stops a merge while it runs,
and it takes about twenty minutes.

## What does not fix it

Resolving the source to the head branch instead of the merge ref, which was the
first attempt in #591 and is now closed. The head branch is deleted on merge as
well, so the same race produces the same failure.

Building from the mounted checkout is forbidden by the packaging contract:
unpublished candidate sourcing is local-only and must stay out of CI. Pointing
the SteamOS checkout somewhere other than the resolve-source output is forbidden
by `scripts/check-release-package-dependencies.py`, which requires all six
packaging jobs to check out the immutable resolved commit.

## Options

1. **Do not merge while the packaging jobs are running.** No code change. It is
   entirely a discipline problem, and the discipline was mine.
2. **Add `SteamOS 3.8 build` to the required checks.** GitHub then refuses the
   merge until the job finishes. It is skipped for web-only pull requests, and a
   skipped required check counts as satisfied, so web work is unaffected. The
   cost is waiting on a twenty-minute job before merging anything native.
3. **Stop deleting the branch on merge.** Only helps in combination with the
   head-branch resolution that #591 proposed, and leaves stale branches behind.

Option 1 costs nothing and fixes it today. Option 2 is the version that does not
rely on anyone remembering.

papi-ux:
## Summary

`SteamOS 3.8 build` has failed on every pull request for weeks while passing on
every push to master, always with the same message, never with anything that
looked like a real error:

```
  -> Creating working copy of polaris git repo...
fatal: unable to read tree (9dc7168134a16bbdd4b2c01ed342c2d6bc11d759)
```

That object is not corrupt and it is not a tree. It is the synthetic merge
commit GitHub builds for a pull request. `git ls-remote` proves the point:

```
$ git ls-remote <repo> | grep 9dc7168134a16bbdd4b2c01ed342c2d6bc11d759
   (nothing: no clone can ever see it)
$ git ls-remote <repo> | grep 7761c6a7          # an open pull request head
7761c6a7...  refs/heads/docs/console-hints-to-docs
7761c6a7...  refs/pull/589/head
```

The SteamOS job is the only one that does not build purely from its own
checkout. Its PKGBUILD sources from a clone of this repository:

```
source=("$pkgname::git+@GITHUB_CLONE_URL@#commit=${_commit}")
```

A clone fetches `refs/heads` and `refs/tags`. The merge commit is in neither, so
makepkg aborts. On a push the commit is an ordinary branch commit and the job
passes.

`POLARIS_CHECKOUT_REF` now resolves to the pull request's head branch instead of
the merge ref. The head is a published commit every clone can find. Because
master requires branches to be up to date before merging, the head already
carries master, so the merge commit was buying no coverage that the head does
not already give.

A pull request from a fork has no commit on a branch of this repository, so the
job is skipped there rather than failing in a way it cannot fix.

## What this deliberately does not do

Two earlier attempts were wrong, and the repository's own guards caught both:

- Building pull requests from the mounted checkout, which is what the Arch job
  does, is forbidden by the packaging contract: unpublished candidate sourcing
  is local-only and must be absent from CI.
- Pointing the SteamOS checkout at the head commit directly is forbidden by the
  release package guard: all six packaging jobs must check out the immutable
  `resolve-source` output and nothing else.

Both rules survive here untouched. `resolve-source` stays the single authority,
the SteamOS job body is byte for byte what it was, and the only behavioural
change is which published ref the whole workflow resolves.

## Exact candidate

Branch `ci/steamos-pr-source`, one commit on master `838e8318`. One `env` entry
in `.github/workflows/build.yml`, one `if` on the SteamOS job, and the contract
pins for both rules.

## Verification

- `python3 scripts/check-release-package-dependencies.py` passes, which is the
  guard that rejected the second attempt.
- 662 web tests across 83 files pass, including `packaging-contracts.test.js`,
  which is the guard that rejected the first attempt. It now pins the published
  source ref and the fork exclusion, so neither wrong approach can come back.
- The mechanism was confirmed against the live remote with `git ls-remote`, as
  quoted above: the merge commit is advertised on no ref a clone fetches, and an
  open pull request head is advertised as a real branch.
- This pull request edits a non-web path, so it is classified as a native build
  and runs the SteamOS job itself. That run is the end-to-end proof.

Closes #590

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,16 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 env:
-  POLARIS_CHECKOUT_REF: ${{ inputs.release_tag || github.ref }}
+  # On a pull request github.ref is refs/pull/<n>/merge, and that merge commit
+  # is published nowhere: it is advertised only in GitHub's pull request refs,
+  # so a clone of this repository cannot resolve it and the SteamOS package
+  # build, which sources from such a clone, dies reading its tree. Resolve the
+  # head branch instead, which is a published commit. Master requires branches
+  # to be up to date before merging, so the head already carries master.
+  POLARIS_CHECKOUT_REF: ${{ inputs.release_tag
+    || (github.event.pull_request.head.repo.full_name == github.repository
+    && format('refs/heads/{0}', github.head_ref))
+    || github.ref }}
   POLARIS_PACKAGE_REF_NAME: ${{ inputs.release_tag || github.ref_name }}
   # Advance deliberately only after the full matrix proves a newer snapshot.
   UBUNTU_APT_SNAPSHOT: 20260901T000000Z
@@ -785,7 +794,14 @@ jobs:
   steamos-build:
     name: SteamOS 3.8 build
     needs: [resolve-source, web-checks, cpp-sanitizer-tests, arch-build]
-    if: needs.resolve-source.outputs.native_required == 'true'
+    # The package is built from a clone of the published repository, so the
+    # commit under test has to be published here. A pull request opened from a
+    # fork has no commit on a branch of this repository, and CI never packages
+    # an unpublished tree, so the job cannot speak for a fork.
+    if: >-
+      needs.resolve-source.outputs.native_required == 'true'
+      && (github.event_name != 'pull_request'
+      || github.event.pull_request.head.repo.full_name == github.repository)
     runs-on: ubuntu-24.04
 
     steps:
--- a/src_assets/common/assets/web/packaging-contracts.test.js
+++ b/src_assets/common/assets/web/packaging-contracts.test.js
@@ -726,6 +726,17 @@ describe('Linux packaging contracts', () => {
 
     expect(steamOs).toContain('git fetch --no-tags --force origin "$POLARIS_CHECKOUT_REF"')
     expect(steamOs).toContain("git rev-parse 'FETCH_HEAD^{commit}'")
+    expect(steamOs).toContain('test "$(git rev-parse HEAD)" = "$POLARIS_BUILD_COMMIT"')
+    // A fork's head commit is published nowhere in this repository, and CI never
+    // packages an unpublished tree, so the job must not claim to speak for one.
+    expect(steamOs).toContain(
+      'github.event.pull_request.head.repo.full_name == github.repository',
+    )
+    // Every package must come from a published commit. On a pull request that is
+    // the head branch, never GitHub's merge commit, which no clone can resolve.
+    expect(workflow).toContain(
+      "format('refs/heads/{0}', github.head_ref)",
+    )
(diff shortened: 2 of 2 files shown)


papi-ux:
Closing this without merging. The diagnosis behind it was wrong, and the fix
would not have prevented a single one of the failures it claimed to fix.

**What actually happens.** GitHub advertises `refs/pull/<n>/merge` for as long as
the pull request is open, so a clone can resolve the merge commit perfectly well.
The ref disappears when the pull request is merged. The SteamOS job clones and
checks out its source several minutes into the run, after the chroot is built and
packages are installed, so merging the pull request while that job is still
running pulls the ref out from under it.

The timings say it plainly:

| PR | SteamOS clone | merged | result |
| --- | --- | --- | --- |
| #588 | 03:45:48 | 03:42:06 | clone ran 3m42s after the merge, failed |
| #589 | 04:46:02 | 04:53:23 | clone ran 7m21s before the merge, passed |

#584 and #586 match #588: each was merged about one minute after its SteamOS job
started, which is well before the clone step. #589 is the control, and it passed
on the unmodified workflow.

So the job was never broken for pull requests. It was broken by merging on the
four required checks while the twenty-minute packaging job was still running,
which is exactly what I did on #584, #586, and #588.

**Why this change does not help.** It resolves the source to `refs/heads/<head>`
instead of the merge ref, but the head branch is deleted on merge too, so the
same race produces the same failure. It would also have changed what every job
builds on every pull request, for no benefit.

The real options are in #590, which I have corrected.

PR #594, opened 2026-09-03
docs(changelog): record the seat isolation and Heroic runtime changes

papi-ux:
## Summary

Two changes landed on master after the last changelog edit and are missing from
`Unreleased`, so a 1.4.1 cut today would ship them unrecorded:

- **#592**, which makes Polaris say when client keyboard and mouse seat isolation
  did not take effect, instead of leaving a setting that saves, reads back as
  enabled, and quietly leaves the devices on the desktop session's seat.
- **#593**, which carries the platform and runtime of imported Heroic titles
  through the scan and the import response, so a Windows title under Proton is no
  longer indistinguishable from a Linux native build until it launches.

It also rejoins the `Unreleased` list, which an earlier edit of mine had split in
two with a stray blank line.

This deliberately touches nothing else. The version in `CMakeLists.txt`, the
release notes, and the README version references are all part of cutting the
release itself, and are being handled separately.

## Exact candidate

Branch `docs/changelog-592-593`, one commit on master `727bc80a`, one file.

## Verification

`scripts/check-public-surface.sh` clean. Documentation only, so no build or test
behaviour changes.


papi-ux:
Superseded by the consolidated v1.4.1 candidate in #595. Its release notes include both the #592 seat-isolation warning and #593 Heroic platform/runtime work. Keeping this branch intact until the release candidate is resolved.

PR #648, opened 2026-09-11
Carry a worker's negotiated media to its own client

papi-ux:
## Summary

The controller side of container multiseat's media path. A worker can now
announce what it produces, have that held against what the client negotiated,
and have its frames carried to that client. What is missing is the producer:
no worker announces a contract yet, so nothing takes this path and production
stays exactly as inert as it was.

Before this, a session bound to a worker could not stream at all. The start
path refused it, because the reservation existed but nothing read media from
it, and falling back to host capture would have handed one client another
client's screen.

Six commits, each green on its own:

1. The contract itself: five messages and a per frame prefix carrying the
   index, the keyframe flag and the capture and encode timestamps. Everything
   is fixed size, big endian and validated on both sides, and the C++ and Go
   implementations pin the same golden hex vectors so they cannot drift.
2. The transport: the contract rides the media channel, ahead of the frames it
   describes, so it can never be read out of order with them. The
   acknowledgement, keyframe requests and reference invalidations ride control
   as ordinary requests with one shared acknowledgement, and all four are
   refused until both channels are attached.
3. The host pump: attaches, takes the contract, refuses anything the client did
   not negotiate, acknowledges it, then delivers every frame to the stream's own
   packet destination. From there the ordinary broadcast path does the rest, so
   worker media is sent, paced, encrypted and accounted exactly like the host's
   own. A stopping session closes its transport rather than waiting on a worker.
4. An end to end proof: a real Go worker holds its frames until the native pump
   acknowledges its contract, so the frames arriving is itself the proof that
   the negotiation happened.
5. Documentation: what the media path does now, plus two requirements that were
   nowhere. A profile volume must be mode 0700, and the physical image reference
   must be a digest.
6. The seat's data plane: the rule that a worker sends nothing before its
   contract is acknowledged now lives in production code rather than only in a
   test fixture.

Nothing falls back. A contract the client did not negotiate, a frame before the
contract, a second contract, a malformed frame and a lost transport all end the
stream. The start path refuses only what it should: a session that requires a
worker and does not have a live one.

## Exact candidate

`2b73265` on `feat/multiseat-continuous-media`, based on master `ade0e282`.

## Verification

On an RTX 4090 Fedora 44 host: full build of every target, all 25 CTest suites
(113 s), the Go worker suite, and the public docs and public surface checks.
The new suites are seven media pump cases, four protocol cases pinned against
the Go vectors, one transport case, two worker contract cases, six seat data
plane cases, and the end to end interop case. `git diff --check` is clean.

The physical multiseat harness also passed on that host for the first time,
against master rather than this branch: two isolated Gamescope workers, input
reaching only the targeted seat in both directions, one seat still running and
still receiving input after the other stopped, per seat Opus observations on
both seats, and nothing left behind.

Not verified here: the sanitizer suites, which is why this is open. The pump
runs a reader and a request thread against a shared connection, and TSan has
not seen it yet.

## Not a merge candidate yet

The encoder provider is the one thing between this and a worker that streams.
Activation is deliberately not included: a config key that starts seats which
announce nothing would only produce sessions that end themselves.


papi-ux:
Rebased onto current master and re-verified locally. Landing this is debt avoidance, not feature delivery, and the gate that actually blocks the arc is named below so nobody has to rediscover it.

**Rebase.** Eight commits replayed onto `master` with zero conflicts. The two changed-file sets are disjoint: this branch touches `src/stream.cpp` and four files under `src/platform/linux/`, while master since the base touched `src/process.cpp`, `src/video.cpp` and `stream_display_policy.*`. Master's new tree-wide source contracts were checked against this branch's 3300 new lines: no `std::system`, no `popen`, no locale-sensitive decimal parsing.

**Re-verified on the rebased tree:** full `ctest -j 8` at 25 of 25, the Go worker suite clean, and the live host config byte-identical before and after.

**What merging buys: nothing a user can reach.** Every line is behind three closed gates, and that is deliberate:

- The worker's production entrypoint passes `nil` adapters and `nil` data plane, so `newSeatDataPlane` has no production caller at all.
- The host flag `multiseat_moonlight_input` defaults false, and even flipped on it injects only an input backend, never a worker authority.
- The stream path refuses a worker session whose worker connection is not live.

**What actually blocks the arc, in order.** None of it is in this PR:

1. **The encode-observation defect.** A sixty-frame probe attaches to a seat's raw-frame socket and takes the producer down with it, killing the compositor and the game on both seats. Reproduced on the physical host; passes with observations disabled. This is small, one to three days, and strictly first, because the encoder provider will be a permanent consumer of that same socket and will hit the same class of bug on day one.
2. **The encoder provider.** The `mediaSource` seam has exactly one implementor and it is a test fixture. There is no encoder binary in the worker image, and the image contract test deliberately asserts the provider catalog is absent. One and a half to three weeks, including a worker image rebuild across the locked app roots.
3. **The activation key.** Small, and correctly last: a key that starts seats which announce nothing would only produce sessions that end themselves.

Realistically three to five weeks before a client sees a containerised seat's frames. Merging now simply stops 3300 reviewed, sanitizer-clean, SteamOS-clean lines from decaying against a master that moved ten commits in a day.

PR #673, opened 2026-09-12
fix(input): give a client the controller it asked for, from its second launch on

### LINKED ISSUE #671: Controller type the client asks for is discarded on every launch
## what happens

on every Linux launch that has a command, Polaris creates the virtual gamepad **before** the client
has told it what kind of pad to emulate, and then throws the client's answer away when it arrives.

the result is that auto-detection never runs. with the default `gamepad = auto`, a client with a
DualSense gets an Xbox One pad, and loses the touchpad, the motion sensors and the adaptive
triggers. the only workaround is the host-global `gamepad = ds5`, which is not per client, so it is
wrong for anyone else who connects.

there is some irony in it. #660 fixed the virtual DualSense mapping and resting axes, and on this
path nobody on `auto` ever reaches that pad.

## why

`input::preallocate_gamepad()` allocates controller 0 with an empty arrival:

```cpp
// src/input.cpp:275
platf::alloc_gamepad(platf_input, {id, controller_number}, {}, preallocated_gamepad_feedback_queue)
```

so `metadata.type` is `LI_CTYPE_UNKNOWN` and `metadata.capabilities` is 0. the type ladder in
`src/platform/linux/input/inputtino_gamepad.cpp:119-155` then has every auto branch unreachable:

```
:134  metadata.type == LI_CTYPE_XBOX       unreachable
:137  metadata.type == LI_CTYPE_PS         unreachable
:140  metadata.type == LI_CTYPE_NINTENDO   unreachable
:143  motion sensors present               unreachable, capabilities == 0
:146  touchpad present                     unreachable, capabilities == 0
:149  else -> Xbox One (default)           always taken
```

the client's real `SS_CONTROLLER_ARRIVAL` then arrives and is dropped, because the pad already
exists:

```cpp
// src/input.cpp:963-965
if (input->gamepads[packet->controllerNumber].id >= 0) {
  BOOST_LOG(debug) << "ControllerNumber already allocated [" << ... << ']';
  return;
}
```

that is a `debug` line, so nothing in a normal log says the client's declared type was ignored. the
capability-mismatch warnings at `inputtino_gamepad.cpp:154-171` are silenced for the same reason,
since they all test `metadata.capabilities`.

preallocation is gated on `has_launch_commands` alone (`src/process.cpp:8840,8949`), not on
headless or cage, so this is every Linux launch with a command, not just Private Stream.

## evidence

a reporter's log on #634, on a client that would otherwise have been auto-detected:

```
Info: ControllerNumber [0] preallocated before app launch
Info: Gamepad 0 will be Xbox One controller (default)
```

`(default)` is the tell. the auto-selected variants of that line are never printed on this path.

## why it is not a one line fix

the pad has to exist before the app starts, so Steam and friends see it at startup. the client's
arrival only comes once the stream is up, which is after that. and under strict isolation the
bubblewrap namespace binds the node list at exec time
(`src/platform/linux/input/inputtino_gamepad_isolation.cpp:566-631`), so a pad destroyed and
re-created after launch is invisible to the app: `--tmpfs /dev/input` plus explicit `--dev-bind`
entries resolved once, and new kernel nodes do not appear behind an existing tmpfs.

nothing remembers a client's pad type between sessions either. not `client_profiles`, not
`device_db`, not `named_cert_t`. the type is only ever sent in the in-stream arrival packet.

## proposed fix

remember the controller type per paired client and preallocate with it. the first session from a
new device is Xbox as it is today, and every session after that is right. `named_cert_t` already
carries per-client `display_mode` and `target_bitrate_kbps`, so there is precedent for the store.

and raise that dropped-arrival log above `debug`, so when it does happen it is visible.

papi-ux:
## Summary

Closes #671.

- A paired client now remembers the controller type it last declared, next to the display mode
  and bitrate it already carried.
- The pad preallocated before an app launches starts as that type instead of always Xbox One.
- A controller arrival that lands too late to be honoured is recorded for next time rather than
  only dropped.

> **Stacked on #672.** Its base is `polaris/hdr-doctor-finding` because both commits touch
> `stream_stats`. It retargets to master on its own once #672 merges.

Polaris creates the virtual gamepad before the app starts, because a game that enumerates
controllers at startup has to find one there. The client only says which pad it wants once the
stream is up, which is later. That gap was closed by preallocating with an empty arrival
(`input.cpp:275`), so `type` was `LI_CTYPE_UNKNOWN` and `capabilities` were zero, every
auto-detect branch of the ladder at `inputtino_gamepad.cpp:134-148` was unreachable, and the
answer was always the Xbox One default at `:149`. The client's real arrival then landed and was
dropped at `input.cpp:963-965` behind a `debug` line nobody reads.

So on the default `gamepad = auto` a DualSense client got an Xbox pad and lost its touchpad, its
motion sensors and its adaptive triggers, with nothing in a normal log to say why. The
host-global `gamepad = ds5` was the only way out, and it is wrong for everyone else who connects.
Preallocation is gated on `has_launch_commands` alone (`process.cpp:8840`), so this is every Linux
launch that has a command, not only Private Stream.

The pad cannot simply be swapped once the arrival lands. Under strict isolation bubblewrap binds
the device node list when it execs (`inputtino_gamepad_isolation.cpp:566-631`), so a pad destroyed
and re-created after the app started is invisible to it. What is possible is remembering. A device
streaming for the first time still gets Xbox, which is the same as today and the only honest
answer when nothing has been observed; every session after that is right.

The route stays inside the existing layering. `input.cpp` records what the client declared through
`stream_stats`, which it already depends on. `stream.cpp` persists it at teardown, where the client
uuid already lives and where temporary authorizations are already expired. `nvhttp` carries it onto
the launch session from the paired cert, the same way it already carries `device_name`, so
`process.cpp` needs nothing new to reach it.

## Exact candidate

`Commit: adcaaeee`
`Parent: f5ec7315d366d2c7cb9a74908870891bd1b324b0`
`Base: polaris/hdr-doctor-finding`

## Verification

- [x] `ctest -j 8` on the CUDA-off tree: 25 of 25 suites passed.
- [x] `PairingAccessPresetTest.RemembersTheControllerTypeAClientDeclared` covers the store: a
      never-observed device is left alone, a first answer is kept, the same answer again is
      refused so a disconnect does not churn the state file, a changed answer is kept, and an
      unknown uuid is refused.
- [x] Two source guards in `tests/unit/test_preallocated_gamepad_source.cpp`, since creating a
      uinput device is not something every runner can do. They pin that preallocation does not
      pass an empty arrival again, and that the late arrival is recorded before the early return.
- [x] **RED proven, one piece at a time.** Restoring the empty `{}` arrival fails the first guard;
      removing the late-arrival record fails the second; making the store always refuse fails the
      pairing test. Green again when each is restored.

## What is not covered

No hardware run. The type ladder in `inputtino_gamepad.cpp` is not exercised by any test in this
repo today and still is not, so what is proven is that the remembered type is stored, carried onto
the launch session and handed to preallocation, not that a DualSense appears on a second launch.
That wants a device and is the obvious next check.

The first session from any new device still gets an Xbox pad. That is unchanged behaviour rather
than a fix, and a client that never sends an arrival never teaches the host anything.

### DIFF
--- a/src/crypto.h
+++ b/src/crypto.h
@@ -107,6 +107,10 @@ namespace crypto {
     std::string uuid;
     std::string cert;
     std::string client_family;
+    /// Last controller type this client declared on arrival (an LI_CTYPE_* value), so the
+    /// pad preallocated before the app launches can be the one it actually asked for.
+    /// Zero means never observed, which is what a first session from a new device looks like.
+    int controller_type = 0;
     std::string display_mode;
     int target_bitrate_kbps = 0;
     std::int64_t paired_at = 0;
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -251,7 +251,7 @@ namespace input {
     return true;
   }
 
-  void preallocate_gamepad() {
+  void preallocate_gamepad(int client_controller_type) {
     if (!config::input.controller) {
       return;
     }
@@ -272,7 +272,12 @@ namespace input {
     }
 
     constexpr int controller_number = 0;
-    if (platf::alloc_gamepad(platf_input, {id, static_cast<std::uint8_t>(controller_number)}, {}, preallocated_gamepad_feedback_queue)) {
+    const platf::gamepad_arrival_t remembered {
+      static_cast<std::uint8_t>(client_controller_type),
+      0,
+      0
+    };
+    if (platf::alloc_gamepad(platf_input, {id, static_cast<std::uint8_t>(controller_number)}, remembered, preallocated_gamepad_feedback_queue)) {
       free_id(gamepadMask, id);
       update_controller_diagnostics(false, controller_number, "ControllerNumber [0] could not be preallocated before app launch.");
       BOOST_LOG(warning) << "ControllerNumber [0] could not be preallocated before app launch"sv;
@@ -961,6 +966,11 @@ namespace input {
     }
 
     if (input->gamepads[packet->controllerNumber].id >= 0) {
+      // The pad was preallocated before the app launched, which is the only way a game sees a
+      // controller at startup, so this arrival is too late to change it. Remember what the
+      // client asked for instead, and say so, because silently emulating the wrong pad for a
+      // whole session used to leave nothing in the log at all.
+      stream_stats::update_client_declared_controller_type(packet->type);
       BOOST_LOG(debug) << "ControllerNumber already allocated ["sv << packet->controllerNumber << ']';
       return;
     }
--- a/src/input.h
+++ b/src/input.h
@@ -25,7 +25,14 @@ namespace input {
 
   bool probe_gamepads();
 
-  void preallocate_gamepad();
+  /**
+   * @brief Create controller 0 before the app launches.
+   * @param client_controller_type An LI_CTYPE_* the client declared on a previous session,
+   *        or 0 when nothing is known. The pad has to exist before the app starts so the
+   *        game sees it at startup, which is earlier than the client's own arrival packet,
+   *        so the only way to get the right pad is to remember the last one.
+   */
+  void preallocate_gamepad(int client_controller_type = 0);
   std::shared_ptr<input_t> alloc(safe::mail_t mail);
 
 #ifdef POLARIS_TESTS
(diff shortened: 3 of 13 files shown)


papi-ux:
superseded by #675. GitHub closed this automatically when its base branch `polaris/hdr-doctor-finding` was deleted on merging #672. same commit, same branch, retargeted at master.

PR #719, opened 2026-09-17
Feature/amd encoder profiles

coreydeli:
## Summary

Adds real encoder-profile options for AMD hosts on the two hardware paths Polaris uses there — VA-API (the default for AMD/Intel Mesa) and Vulkan Video (experimental, strict) — replacing a few checkboxes with controls that reach the driver. Verified on an RX 7900 XTX / navi31 RDNA3 under CachyOS + KDE Plasma Wayland; branch synced through upstream master `ae4760dc` (v1.4.10 + #738).

- **Vulkan quality level (`vk_quality`).** New config key (default `0`) wired into all three Vulkan codec tables in `video.cpp` (H.264, HEVC, AV1). Level 0 is the driver default and always works; higher levels trade encode speed for quality where the driver exposes them. Valid values run 0 through one less than the driver's reported maximum — the Vulkan spec requires `qualityLevel < maxQualityLevels` (`VUID-VkVideoEncodeQualityLevelInfoKHR-qualityLevel-08311`, `VUID-VkPhysicalDeviceVideoEncodeQualityLevelInfoKHR-qualityLevel-08261`) while FFmpeg only checks `quality > maxQualityLevels` — so on an RX 7900 XTX, where verbose logging says `Encoder max quality: 4`, the valid levels are 0–3. The select offers exactly that range; a level above it fails that codec's validation with FFmpeg's exact error rather than falling back, since an explicit Vulkan selection is strict.
- **Hardware codec support panel.** New read-only `CodecSupportPanel.vue` on the VA-API and Vulkan tabs: active encoder plus H.264/HEVC/AV1 rows, marked HDR where the probe accepted a Main10/P010 configuration, with AV1-preferred / HEVC-fallback guidance. It is labeled as what Polaris advertises to clients and says why a codec is off — disabled in configuration (mode 1, never probed) or not available on this encoder (e.g. Vulkan's `NO_AV1`) — with the reasons computed in C++ from probe result plus config state and served as flat keys (`hevc_reason`, `av1_reason`) beside the existing booleans. Served by a new **response-only** `encoder_codec_support` key on the config API; registered in `confighttp_validation.cpp`, which also gains `vk_quality`. The panel refreshes with the other host capabilities after a restart from console or tray (`ConfigView.vue`).
- **`vk_rc_mode = 0` (auto) sentinel fix.** In all three Vulkan codec tables, raw `0` was passed as a rate-control flag and selected the driver's default mode instead of FFmpeg's auto resolution. Auto now passes the named `"auto"` constant; explicit modes are still passed to the driver as-is. Because Polaris always sets a stream bitrate, auto resolves to VBR when the driver advertises it and CBR otherwise — noted in docs, `en.json` and the changelog so anyone already on 0 knows what changes for them.
- **VA-API tab styling repair.** Upstream's `VAAPIEncoder.vue` used undefined `form-*` classes; the three selects now use Polaris conventions (the same label/input treatment every other encoder tab uses) and the component is added to the settings-affordances test sources.
- **Tests + docs.** Two new gtests cover the rc-mode→FFmpeg mapping in `test_video.cpp`; a 6-test vitest suite covers `CodecSupportPanel.vue`; the Vulkan encoder contract test pins the `vk_quality` select (id, default level, and the 0–3 cap). `docs/configuration.md` gains a VA-API Encoder section and documents `vk_tune`, `vk_rc_mode` and `vk_quality`, including how to read the quality range from the log; changelog entries under `## Unreleased`.

## Testing

- [x] I updated docs or release notes when user-facing behavior changed.

All at head `ca28eafe` (synced through upstream master `ae4760dc`), on the RX 7900 XTX / navi31 test host:

| Check | Result |
|---|---|
| C++ build (`-DBUILD_TESTS=ON -DBUILD_FULL_TESTS=ON`) | ✅ exit 0, re-run at `ca28eafe` |
| ctest (25 suites) | ⚠️ 24/25 — one pre-existing local-env failure, explained below; passes in CI |
| Web lint + vitest | ✅ clean; **1024/1024** across 120 files |
| VA-API default (auto/auto) | ✅ `encoder_codec_support` reports AV1 + HEVC supported with HDR; auto policy picks CBR/VBR per codec; clean session settings and encoder open |
| VA-API explicit `vaapi_rc = cbr`, `vaapi_quality = balanced` | ✅ both keys round-trip via the API; CBR on all codecs |
| VA-API unsupported RC (`vaapi_rc = icq`) | ✅ graceful: warning → auto policy → encoder opens, no crash (navi31 supports CQP/CBR/VBR/QVBR) |
| Vulkan `vk_quality` 0–3 | ✅ accepted; log shows `Encoder max quality: 4`; cache saved |
| Codec support panel reasons | ✅ disabled-in-config and not-available-on-encoder states render the right reason on both tabs (vitest + live API check) |

The live stream rows were exercised on this host during validation at `7c2b4d9f`; nothing in the encoder path changed after it. The one ctest failure is `PortalCapabilityPolicyTests.PreparationLeavesProcRootReadableToSameUserPortalPeer` in `test_polaris_platform`: this machine's user session carries an inheritable `cap_wake_alarm`, and upstream's heuristic in `portal_capability.cpp` treats any non-zero capability set as "must drop", which requires a single-threaded process that gtest never is. It fails on upstream master too, in isolation, on this box; CI (no inheritable caps) takes the unchanged branch and passes. Unrelated to this PR — the diff touches nothing under `src/platform/`.

## Notes

- No changes to pairing, trusted subnets, certificates, client commands, shell hooks, packaging scripts, or privilege boundaries.
- The `encoder_codec_support` key is response-only by design: it is on the response-only list and `apply_config` skips it, so it cannot leak into config writes; Nova advertisement is unchanged.
- One small extra beyond the review asks: I also updated the `vk_rc_mode_desc` copy in `en.json` to mention that auto resolves to VBR (then CBR) when a bitrate is set — the console tooltip carried the same stale wording as the docs. Easy to drop if you'd rather keep it out.


coreydeli:
I have validated this by making a build locally and running it. AV1 support functions fully at this point as well as first class AMD support using my 7900XTX as a testing ground.

papi-ux (commented):
hey @coreydeli, thank you for this! the `vk_rc_mode` auto fix is a real bug (a raw 0 was selecting the driver's own mode instead of FFmpeg's auto) and the VA-API tab cleanup was overdue, so i really appreciate you digging into both. i went through the whole diff against the FFmpeg we pin and there are a few things i want to sort out before we merge.

### needs a change before merge

**1. the `vk_quality` range is off by one**

In the Vulkan spec `maxQualityLevels` is a count, so the valid levels are 0 through `maxQualityLevels` minus one. The FFmpeg we pin (release/8.1 at `38b88335`) checks `quality > maxQualityLevels` in `libavcodec/vulkan_encode.c:900`, so it lets one extra level through without an error. On your 7900 XTX, where the log says `Encoder max quality: 4`, the valid levels are 0 to 3, and level 4 opening cleanly isnt proof that its a real level.

Right now:
- `docs/configuration.md` documents the range as `0..maxQualityLevels` and says the 7900 XTX maximum is 4
- `VulkanEncoder.vue:56` offers levels 1 to 8, so on that card 5 through 8 fail the encoder open and 4 is already out of range

could you document the range as 0 through `maxQualityLevels` minus one (and update `vk_quality_desc` to match), and either cap the select or show the maximum the probe found?

**2. the codec support panel shows what Polaris advertises, not what the hardware passed**

`encoder_codec_support` reads `hevc_mode` and `av1_mode >= 2` (`confighttp.cpp:5173`), which is the advertised policy after the probe. A few cases where that reads wrong:
- if `hevc_mode` or `av1_mode` is set to 1, that codec is never probed (`video.cpp:4918`), but the panel says "Not supported"
- the Vulkan encoder carries `NO_AV1`, so on the Vulkan tab AV1 always shows "Not supported", even on a card that can encode AV1, right next to the note about hardware validation
- both tabs show the same data, for whichever encoder is active

either expose the probe result per codec, or label the panel as what gets advertised to clients and say why a codec is off (disabled in config, not available on this encoder). either works for me.

**3. repo hygiene**
- please drop `.codegraph/.gitignore` and the `.tmp/` and `docs/qwen.md` lines in `.gitignore`, since CONTRIBUTING asks to keep local workspace files out
- the PR description reads like a working doc (Part A and Part B). could you swap it for the template (Summary, Testing, Notes) and move the Part B bullets into `docs/changelog.md` under `## Unreleased`?

### should change

- the panel never refreshes: `refreshHostCapabilities` in `ConfigView.vue` only re-reads a few keys, so "supported codecs appear here once probing completes" isnt true until a page reload. adding `encoder_codec_support` there and adjusting that copy would fix it
- the `vk_rc_mode` docs still say 0 lets FFmpeg and the driver decide. with the auto constant, FFmpeg picks VBR (then CBR) whenever a bitrate is set (`vulkan_encode.c:668`), so anyone already on 0 moves from the driver default to VBR. worth a line in the changelog
- tests: nothing covers `CodecSupportPanel.vue`, the new `vk_quality` select, or the `"auto"` rc_mode string
- the testing table was run on `2f70b50`, and the last master merge brought in a lot (`confighttp.cpp` and `config.cpp` especially), so a rerun on the current head would be great

### small stuff, optional

- the panel's status pills could reuse `components/StatusBadge.vue`
- the same option lambda and comment show up three times in `video.cpp`, one helper would do
- "driver default" for level 0 is a little misleading, since 0 is just the first level, and the new docs section has an em dash in it

### what checks out

- `quality` and `rc_mode` both live in FFmpeg's shared Vulkan option table for H.264, HEVC and AV1, the `"auto"` constant is the right one, and a default of 0 changes nothing
- `encoder_codec_support` really is response only: its on the response only list and `apply_config` skips it, so it cant leak into config writes
- the VA-API styling fix is real, those `form-*` classes dont exist anywhere
- adding the new strings only to `en.json` matches how the repo handles locales

i dont have an AMD card in the lab so none of this was run on hardware on my end, thats all from reading the code and the FFmpeg source. thanks again for putting this together, just hit me up if anything here doesnt make sense!

coreydeli:
Looks good overall. I will make the changes and update the PR.

papi-ux:
hey @coreydeli, a couple follow ups after digging a bit further.

on the vk_quality range, i went and checked the spec and i made a mistake. the Vulkan spec calls maxQualityLevels the number of quality levels, and two valid usage rules say qualityLevel must be less than it:

- VUID-VkVideoEncodeQualityLevelInfoKHR-qualityLevel-08311
- VUID-VkPhysicalDeviceVideoEncodeQualityLevelInfoKHR-qualityLevel-08261

so on your card where the log says Encoder max quality: 4, the real levels are 0 to 3, and 4 is out of spec. seemed to have opened cleanly because RADV doesnt range check it and FFmpeg only checks quality > maxQualityLevels, so the one boundary value your table proves works is the one that shouldnt. the validation layers would (and should) have flagged it.

also worth knowing, FFmpeg master still has that same > today, not just the 8.1 we pin. so we cant lean on FFmpeg to reject an out of range level at all, which makes capping the select the thing that actually matters rather than a nice to have. thats a real FFmpeg bug if anyone feels like filing it upstream.

on the merge commits, dont worry about them. i went through the three dot diff against master and all 23 deletions are intentional, nothing upstream got eaten aaand the nanors pin is back in step. ill squash at merge so the six master merges stay out of history. also side note, the one thing squashing wont do is drop .codegraph/.gitignore and the .tmp/ and docs/qwen.md lines, those still need deleting por favor.

everything else from the review stands. take your time and for real thank you so much for helping with this! you're the goat.

coreydeli:
Thanks for the careful pass and for checking the spec — you're right on both counts, and sorry it took a round trip to get the range exactly right. Everything from the review is addressed; the branch has also picked up master since your follow-up (now synced through `ae4760dc`, v1.4.10 + #738), so all of this was re-verified at head `fc74f6fd`.

**Needs change**

1. **`vk_quality` off-by-one.** Agreed — valid levels are 0 through `maxQualityLevels − 1` per `VUID-VkVideoEncodeQualityLevelInfoKHR-qualityLevel-08311` / `VUID-VkPhysicalDeviceVideoEncodeQualityLevelInfoKHR-qualityLevel-08261`, and FFmpeg's `quality > maxQualityLevels` check (still present on master) lets exactly N through. The select is now capped at 0–3 with level 0 labeled "Level 0 (default)", `docs/configuration.md` documents the range as one less than the driver's reported count (with the VUIDs and the RX 7900 XTX data point), and `vk_quality_desc` in en.json matches. Happy to file the FFmpeg bug against ffmpeg/ffmpeg if you want — or it's yours if you'd rather file from your side.
2. **Codec support panel.** Went with your second option: the panel is now labeled as what Polaris advertises to clients, and each codec row says why it's off — *Disabled in configuration* (mode set to 1, never probed) or *Not available on this encoder* (e.g. Vulkan's `NO_AV1` next to a card that can do AV1). The reasons are computed in C++ from the probe result plus config state and served as flat keys (`hevc_reason`, `av1_reason`) beside the existing booleans, so both tabs read correctly for whichever encoder is active.
3. **Repo hygiene.** `.codegraph/.gitignore` deleted; the `.tmp/` and `docs/qwen.md` lines are out of `.gitignore` — it now matches upstream byte-for-byte (verified by diff against master). PR description swapped for the template, Part B bullets moved into `docs/changelog.md` under `## Unreleased`.

**Should change**

- **Panel refresh:** `encoder_codec_support` is in the `refreshHostCapabilities` key loop in `ConfigView.vue`, and the copy now says the panel updates after a restart from console or tray, since changing the encoder changes what Polaris advertises.
- **`vk_rc_mode = 0` behavior change:** noted in docs, en.json and the changelog — auto resolves to VBR when the driver advertises it (CBR otherwise) because Polaris always sets a stream bitrate.
- **Tests:** two new gtests for the rc-mode→FFmpeg mapping (`auto` sentinel + explicit pass-through) in `test_video.cpp`, plus a 6-test vitest suite for `CodecSupportPanel.vue` covering both reason states and the no-key render path.
- **Testing table rerun at `fc74f6fd`:** C++ build clean; ctest 24/25 with one pre-existing local-env failure (`PortalCapabilityPolicyTests.PreparationLeavesProcRootReadableToSameUserPortalPeer` — this box carries an inheritable `cap_wake_alarm` that trips upstream's drop-capabilities heuristic in a multi-threaded gtest process; fails on master too, passes in CI); vitest 1023/1023; lint clean. Full table is in the PR description.

**Small stuff**

- **Pills:** the panel uses the console's `meta-pill` class with the same semantic tone variants that `StatusBadge.vue` renders (and that DashboardView applies inline for its streaming pill) — kept as inline spans because each row mixes supported/HDR/neutral states rather than a single status.
- **Tripled option block:** now one shared `vulkan_common_options` vector used by all three codec tables, which is also what let the rc-mode mapping become a testable free function.
- **"Driver default" label + em dash:** level 0 reads "Level 0 (default)" and the docs section heading no longer has an em dash.

One flag beyond the asks: I also updated `vk_rc_mode_desc` in en.json to carry the VBR→CBR clause — you asked for docs + changelog, but the console tooltip had the same stale wording as the docs. Say the word if you'd rather I drop it.

And thanks — "the goat" is very kind; this one's on you for pointing at the spec VUIDs. The two extra sync merges since your review will disappear with the squash anyway, so nothing to do there.

papi-ux:
hey @coreydeli ! thank you for the super thorough writeup, this is sooo close. quick thing before i dig in though, i think the new commit didnt make it up. the PR head is still a7c7f8dc from last night and fc74f6fd doesnt exist on GitHub i dont think, so the diff still has the 1 to 8 select, .codegraph/.gitignore and the two .gitignore lines, and no changelog, tests or reason keys. any chance you could push it when you get a sec? no need to merge master again, it merges clean and ill squash anyway.

alsoo one thing to look at while youre in there. capping the select at 0 to 3 is right for your 7900, but the count changes per driver and chip. ANV reports 1 level, so on Intel only 0 is valid, and RADV reports whatever the chip says. and since FFmpeg lets a level equal to the count through, the config file or the API can still hand the driver an out of spec level even with the select capped. so i think Polaris needs to clamp on its side: read maxQualityLevels for the codec during the Vulkan probe, clamp to that minus one with a log line, and serve the max in encoder_codec_support so the select only offers real levels. what do you think?

also also side note, keep the vk_rc_mode_desc change, it matches what FFmpeg actually does. and the portal test passes on my Fedora box, so agreed thats your setup and not the PR.

you're amazing thanks again, this is looking really close!

papi-ux:
hey @coreydeli, the push came through, thank you! everything from the first pass is in there: the 0 to 3 select, the reason keys, the clean .gitignore, the changelog and the tests. i merged it onto current master here and it builds and passes the full unit suite and vitest.

two small things left:

- master moved under you with #741, so docs/changelog.md conflicts now, just the two Unreleased lists landing on the same spot. i can resolve that on your branch if thats ok with you, since it's only keeping both lists.
- the clamp from my last comment is still open, and id love for you to take it since its your encoder work: read maxQualityLevels in the Vulkan probe, clamp to that minus one with a log line, and serve the max in encoder_codec_support so the select only offers real levels. no rush, ill hold the merge for it.

thanks again, this one is going to make a lot of AMD folks happy

coreydeli:
hey! pushed — new head is `cc77cdee` (just the clamp commit on top of `ca28eafe`, no master re-merge).

changelog conflict: yes please resolve it on my branch keeping both lists, i'll stay out of that file.

the clamp landed per your spec: during the probe each codec's maxQualityLevels is read from the driver through FFmpeg's own Vulkan device context (same instance + loader function as ff_vk_load_functions), so the count matches exactly what FFmpeg will see at session open — ANV reporting 1 means only level 0 gets offered, RADV reports whatever the chip says. a saved vk_quality above count-1 is clamped to it with a warning log line when the session starts (which also covers the off-by-one where level == count slips past FFmpeg's guard), and encoder_codec_support now serves vk_quality_max so the select only offers real levels — just level 0 until a live probe reports counts.

docs/configuration.md, three unit tests for the clamp, and the web contract test are updated to match; build + unit suite + vitest green at this SHA.

papi-ux:
hey @coreydeli, thank you for the clamp! i merged master into your branch and kept both changelog lists, so CI can finally run. it couldnt start at all while the PR had a conflict, thats why theres nothing on cc77cdee.

two things in the probe i need before we merge though, and the first one is a crash:

**1. the capability query segfaults on RADV.** the spec says an H.264, H.265 or AV1 encode query has to carry that codec's own capabilities struct in the chain (VUID-vkGetPhysicalDeviceVideoCapabilitiesKHR-pVideoProfile-07187, -07188 and -10263), and `query_codec` only chains `VkVideoEncodeCapabilitiesKHR`. RADV writes straight into the codec struct without checking that its there (`radv_video.c`, `ext->flags = ...` right after `vk_find_struct`), so it dereferences null. i ran your exact chain on a Steam Deck (RADV VANGOGH, Mesa 26.0 devel, `RADV_PERFTEST=video_encode`) and it died with SIGSEGV for both H.264 and HEVC. with `VkVideoEncodeH264CapabilitiesKHR` / `VkVideoEncodeH265CapabilitiesKHR` chained after the encode caps it returns `VK_SUCCESS` and `maxQualityLevels` 2. this runs inside Polaris every time it opens the Vulkan device, so on AMD it would take the daemon down. FFmpeg does it the same way (`vulkan_encode.c`, `ctx->enc_caps.pNext = codec_caps`), so matching that is the fix.

**2. the AV1 query doesnt build on Ubuntu 24.04.** noble ships Vulkan headers 1.3.275, which have no `VkVideoEncodeAV1ProfileInfoKHR` (it arrived in 1.3.302), and our Ubuntu build and sanitizer jobs compile against those. wrapping the AV1 block in `#ifdef VK_KHR_video_encode_av1` would do it, AV1 on Vulkan is fail closed anyway.

smaller, whenever youre in there:
- a level saved above what the probe offers (or saved before any probe ran) has no matching option, so the select shows blank. keeping the saved value as its own option would read better
- the vk_quality paragraph in the changelog still says an out of range level fails validation, and now it gets clamped. i left your text as it was when i resolved the conflict

side note, a data point for the docs if you want it: the Deck's VanGogh reports 2 levels for H.264 and HEVC, so only 0 and 1 are real there.

youre awesome, this is really close!

coreydeli:
hey! pushed — new head is `058cf888` on top of your merge, so no conflict to keep this time.

the segfault: you're right and that was on me — the query now chains each codec's own capabilities struct after the generic encode caps (VkVideoEncodeH264CapabilitiesKHR / H265 / AV1), exactly like FFmpeg does in vulkan_encode.c, with the VUIDs noted in a comment so it doesn't regress.

the Ubuntu build: the AV1 query block is wrapped in `#ifdef VK_KHR_video_encode_av1`, so 1.3.275 headers compile it out; AV1 stays fail-closed and -1 leaves it unconstrained. i verified by compiling the file with its real build flags against a header sim that drops the macro — clean.

smaller stuff: a saved level above the probed maximum (or from before any probe ran) now keeps its own option labeled "Quality level N (not offered by this driver)" so the select doesn't read blank, and the changelog paragraph now says out-of-range values get clamped instead of failing validation. also added your VanGogh data point to docs + changelog (two levels → 0–1).

full build + ctest + vitest green at this SHA (the one ctest env failure is the portal-peer one from before).

papi-ux:
thanks, both fixes are right. the query now chains the codec caps the way FFmpeg does, and the AV1 block compiles out on 1.3.275 headers, which is exactly what the sanitizer job tripped on last time.

one thing the master merge did: the six changelog paragraphs ended up under the `v1.4.12` heading, after its closing asset bullet, instead of under `## Unreleased`. the 1.4.12 release contract test catches that, so Web checks would fail on this head. i merged current master (1.4.13 is open now) and moved them back to Unreleased, nothing else touched. pushing that to your branch now and approving the CI runs.

once it's green i'll squash it in.

coreydeli:
Awesome! Sounds good to me! I am going to look into improvements to what can be done on the AMD side now that this part is at the finish line!

PR #746, opened 2026-09-20
fix(video): don't let a dummy-source probe veto HDR on the portal path

### LINKED ISSUE #743: gamescope_stream: dummy-source encoder probe vetoes HEVC/AV1 Main10, host advertises SDR only
### Describe the bug

On the `gamescope_stream` + `capture = portal` path, the encoder probe concludes the encoder cannot do HEVC/AV1 Main10 and disables HDR — but the verdict comes from the *capture device*, not the encoder, and says nothing about what the live path can do.

The portal capture never connects to its PipeWire source while probing:

```
portal: Probe mode — using dummy 1920x1080
```

With no stream there is no DMA-BUF, so `pipewire_dmabuf_negotiated` is false and `make_avcodec_encode_device()` (`portal_grab.cpp:1270`) falls back to the SHM→CUDA device with `prefer_8bit_encode = true`. That device only accepts NV12:

```
cuda::cuda_t doesn't support any format other than AV_PIX_FMT_NV12   (cuda.cpp:142)
```

so the 10-bit probe fails before the encoder is ever asked:

```
Encoder [nvenc] does not support HEVC Main10 on this system
Encoder [nvenc] does not support AV1 Main10 on this system
```

`adjust_encoder_constraints()` then zeroes the mode, which drops the Main10 bits from `ServerCodecModeSupport` (`0x30301` → `0x10101` = H.264 + HEVC8 + AV1 Main8). The client refuses with *"your host PC doesn't support HDR streaming"* before it ever connects.

**The live path is genuinely capable.** A real session negotiates a 10-bit DMA-BUF and feeds P010 through `cuda_dmabuf_t`, which accepts it. Confirmed on this host once the veto was bypassed:

```
portal: dmabuf_negotiate ... negotiated=true
Using frame converter [avcodec_encode_device] target_device=cuda target_residency=gpu target_format=p010
portal: capture_transport=dmabuf frame_residency=gpu frame_format=p010
CUDA DMABUF: fourcc=0x30334258 src_xb30=true src_rgb8=false dst_p010=true
gamescope: pipewire: negotiated a 10-bit stream — the composite is encoded as BT.2020 + PQ (HDR10)
```

Client reported `Video stream: 2880x1800 120 FPS (Codec: HEVC 10-bit HDR)`.

Two follow-on problems:

1. **The wrong verdict is persisted.** Probing while the capture target happened to be restarting wrote `hevc_mode = 1` into `polaris.conf`, permanently disabling HDR until edited by hand.
2. **Fixing the advertisement alone is not enough.** `make_encode_session()` gates the live stream on the same per-codec flag, so a session still dies at launch with a black screen and *"No video received from host"*:

```
hevc_nvenc: dynamic range not supported     (video.cpp:3285)
```

### Steps to reproduce

1. Host with NVIDIA + CUDA, `linux_stream_mode = gamescope_stream`, `capture = portal`, `hevc_mode = 3`.
2. Start Polaris and read the encoder probe in `polaris.log`.
3. `curl -s "http://127.0.0.1:47989/serverinfo?uuid=test" | grep ServerCodecModeSupport` → `65793` (`0x10101`), no Main10 bits.
4. Request HDR from the client → refused before connecting.

### Install method

Local package build (Nix flake from the repo's own `nix/` module).

### Host OS

NixOS 26.11

### GPU and driver

NVIDIA RTX 4070 Ti Super, driver 610.57.04

### Runtime path

Headless (gamescope_stream)

### Expected behavior

A probe that cannot reach the transport the live path will use should not veto Main10. Either treat the verdict as *unknown* rather than *false* in that case, or give the probe a real 10-bit source.

I have a patch that takes the first approach (keeps the configured mode and sets the per-codec flag when the probe ran against the portal's dummy source) and am happy to open it as a PR.

kubicgruenfeld:
## Summary

Fixes #743.

On `gamescope_stream` + `capture = portal`, the encoder probe decides HEVC/AV1 Main10 is unsupported and disables HDR — but the verdict comes from the *capture device*, not the encoder.

The portal capture never connects to its PipeWire source while probing (`portal: Probe mode — using dummy 1920x1080`). With no stream there is no DMA-BUF, so `make_avcodec_encode_device()` falls back to the SHM→CUDA device, which only accepts NV12 (`cuda.cpp:142`). The 10-bit probe therefore fails before the encoder is ever asked, `adjust_encoder_constraints()` zeroes the mode, and the Main10 bits drop out of `ServerCodecModeSupport` (`0x30301` → `0x10101`). Clients refuse with "your host PC doesn't support HDR streaming" before connecting.

The live path is genuinely capable: a real session negotiates a 10-bit DMA-BUF and feeds P010 through `cuda_dmabuf_t`, which accepts it.

This treats the Main10 verdict as **unknown** rather than **false** on the one path where the probe cannot reach the transport the live session will use — portal capture under `gamescope_stream` — and keeps the configured mode. Everything else is untouched: other capture paths, other platforms, and the non-HDR `PASSED` checks all behave exactly as before.

Two call sites need it, which is why there are two commits:

1. `adjust_encoder_constraints()` — otherwise the host advertises SDR only.
2. the per-codec `DYNAMIC_RANGE` flag — `make_encode_session()` gates the live stream on the same flag, so fixing only the advertisement produces a host that advertises HDR and then refuses it at launch with `hevc_nvenc: dynamic range not supported` (`video.cpp:3285`), leaving the client on a black screen with "No video received from host".

When the probe is bypassed the path logs at info rather than silently, e.g.

```
Encoder [nvenc] failed the HEVC Main10 probe against the portal's dummy source;
keeping the configured mode because that probe cannot reach the live 10-bit DMA-BUF path
```

An alternative fix would be to give the probe a real 10-bit source; that seemed a much larger change, and I'm happy to go that way instead if you prefer.

## Testing

Tested on NixOS 26.11, RTX 4070 Ti Super, driver 610.57.04, `gamescope_stream` + `capture = portal`, Moonlight client on a 2880x1800 HDR panel.

Before: `ServerCodecModeSupport` = `65793` (`0x10101`), client refuses HDR.
After: `197377` (`0x30301`), and a real HDR session runs end to end:

```
portal: dmabuf_negotiate ... negotiated=true
Using frame converter [avcodec_encode_device] target_device=cuda target_format=p010
portal: capture_transport=dmabuf frame_residency=gpu frame_format=p010
CUDA DMABUF: fourcc=0x30334258 src_xb30=true src_rgb8=false dst_p010=true
gamescope: pipewire: negotiated a 10-bit stream — BT.2020 + PQ (HDR10)
```

Client reported `Video stream: 2880x1800 120 FPS (Codec: HEVC 10-bit HDR)`. SDR sessions on the same host were unaffected.

Not tested: AMD/VAAPI, and non-portal capture paths (unchanged by this patch, but I have no hardware to confirm).

## Notes

No changes to pairing, trusted subnets, certificates, client commands, shell hooks, packaging scripts, or privilege boundaries.


papi-ux:
the diagnosis here is better than the bug report i would have written. the probe never connects to its pipewire source, so the 10 bit failure is the SHM to CUDA fallback rejecting anything that isnt NV12, not the encoder saying anything at all. i checked the part i was worried about and it is safe: capabilities are fully recomputed on every probe (`capabilities.reset()` in video.cpp), so forcing DYNAMIC_RANGE back on cant leak into a later probe.

one change before i take it. the override fires on `active_hevc_mode == 3 && !encoder->hevc[encoder_t::DYNAMIC_RANGE]` without asking whether 8 bit passed. on a host whose encoder has no HEVC at all, `capabilities.reset()` clears every bit, so DYNAMIC_RANGE is false, this branch sets it to 1, and PASSED stays false. that host then advertises HEVC Main10 and fails at stream start, where today it falls back to h264 cleanly.

the dummy source is NV12, so an 8 bit failure is authoritative, its only the 10 bit verdict that cant be trusted. gating the override on `encoder->hevc[encoder_t::PASSED]`, and the same for av1, keeps the fix and drops that case.

push that and ill merge it. thanks for chasing this one down!

kubicgruenfeld:
Done — pushed as 1449ee57.

Both branches now read `main10_probe_is_authoritative || !encoder->hevc[encoder_t::PASSED]` (and the same for `av1`), so the override only fires when 8-bit genuinely passed. Your no-HEVC host takes the authoritative path again: `capabilities.reset()` leaves `PASSED` false, `active_hevc_mode` goes to 0, and it falls back to H.264 as before.

Comment in the source spells out the reasoning you gave — the dummy source is NV12, so an 8-bit failure is authoritative and only the 10-bit verdict is unreachable.

PR #747, opened 2026-09-20
feat(gamescope): draw the cursor into the captured stream

kubicgruenfeld:
## Summary

On `gamescope_stream` the pointer works but is invisible: input lands, nothing is drawn.

gamescope keeps the cursor out of the PipeWire capture unless `--pipewire-composite-cursor` is passed — reasonably, since a consumer that draws its own pointer would otherwise show two. Moonlight draws none, so it gets none. Nothing in the session scripts passes the flag, so it is always off.

Worth distinguishing from the compositor's own `cursor_composite` convar, which governs the gamescope *output* rather than the *capture node*: setting that to "always composite" changes nothing here, because the capture node is what the client sees. `--pipewire-composite-cursor` is the one that matters, and this passes it for both the idle instance and the nested session, since either can be the one the portal is attached to.

`POLARIS_GAMESCOPE_COMPOSITE_CURSOR=0` restores the previous behaviour for a consumer that does draw its own pointer.

## Testing

Tested on NixOS 26.11, RTX 4070 Ti Super, driver 610.57.04, `gamescope_stream` + `capture = portal`.

Confirmed first at runtime on a live session (`gamescopectl pipewire_composite_cursor 1` made the cursor appear immediately), then with the flag in place across session restarts. Verified the idle instance comes up carrying it:

```
gamescope --backend headless --expose-wayland --steam --xwayland-count 2 --pipewire-composite-cursor -W 3840 -H 2160 …
```

Not tested: a client that renders its own pointer (the double-cursor case the default guards against) — that is what the env var is for.

## Notes

Shell-only change to the two session launch scripts; no changes to pairing, trusted subnets, certificates, client commands, or privilege boundaries.


papi-ux:
nice catch. the distinction you drew between `--pipewire-composite-cursor` and the `cursor_composite` convar is the part i keep seeing people get wrong, the capture node is what the client sees and nothing was passing the flag.

taking it as is. the only thing i thought about was an older gamescope rejecting an unknown flag and failing to start, but `--hdr-enabled` is already passed the same unguarded way a few lines up in the same file, so this follows the assumption thats already there rather than adding a new one.

thank you!

PR #748, opened 2026-09-20
feat(input): route mouse and keyboard into gamescope over libei

kubicgruenfeld:
## Summary

`gamescope_stream` has gamepads only — mouse and keyboard never reach the session.

The one compositor-local input route is labwc-specific: `connect_if_needed()` bails unless `cage_runtime_active()`, which requires `backend_name == "labwc"`. So mouse and keyboard fall through to host uinput, and a headless gamescope has no libinput seat to read those evdev nodes with. Gamepads work because Steam reads `/dev/input` itself, bypassing the compositor entirely.

Repointing the Wayland route at gamescope isn't possible — its wlserver creates no virtual-pointer or virtual-keyboard manager. But it does run an EIS server (`src/InputEmulation.cpp`) advertising pointer, absolute pointer, button, scroll and keyboard, feeding each event straight into wlserver. That is the same path XWayland's XTEST support already takes on our behalf, since XTEST is itself a libei client — so this talks to it directly, dropping two hops and covering native Wayland clients as well as XWayland ones.

Details worth calling out:

- **Keycodes and buttons are evdev on both sides**, so the mapping tables are shared with the Wayland route and nothing is translated.
- **Scroll carries continuous deltas**, so a sub-notch flick survives instead of being rounded to a wheel click.
- **Absolute input is sent as relative deltas** once a previous position exists. `EIS_EVENT_POINTER_MOTION_ABSOLUTE` lands on `wlserver_mousewarp(..., bSynthetic = true)`, which deliberately leaves `bCursorHidden` alone — an absolute-only client would move the pointer without ever revealing the cursor. `EIS_EVENT_POINTER_MOTION` goes to `wlserver_mousemotion()`, which reveals it.
- **Socket discovery:** Polaris is gamescope's parent and so doesn't inherit `LIBEI_SOCKET`. It derives the same name gamescope builds (`"<wayland display>-ei"`), which libei resolves against `XDG_RUNTIME_DIR`. `LIBEI_SOCKET` is still honoured when present.

The route sits behind the same "did the compositor take it" contract as the Wayland one, so every call site keeps falling back to host uinput when neither claims the event.

Two limits I want to be upfront about:

1. **Touch and pen still have no route.** gamescope has `EIS_DEVICE_CAP_TOUCH` commented out (`// Can add this someday if we want it.`), so there is nothing to send them to; they only learn to block the host-uinput fallback. Happy to follow up on the gamescope side if that'd be welcome.
2. **libei documents `ei_setup_backend_socket()` as "primarily useful for testing and debugging"**, preferring per-client fds. It is the only backend gamescope offers (`eis_setup_backend_socket`), and the one XWayland already uses here, so this follows the existing arrangement rather than introducing a new one.

Adds a `libei-1.0` pkg-config dependency; the route compiles out entirely when it is absent, exactly like the existing `xkbcommon` gate for the Wayland route.

## Testing

Tested on NixOS 26.11, RTX 4070 Ti Super, driver 610.57.04, `gamescope_stream` + `capture = portal`, Moonlight client on a 2880x1800 HDR panel, streaming Steam Big Picture and a Proton game.

Confirmed working: pointer motion, buttons, keyboard, scroll, the cursor staying visible, and session resume (which exercises reconnect after the socket goes away). Verified the route is the one carrying input rather than a fallback:

```
EI virtual input: routing mouse and keyboard to the gamescope session on [gamescope-0-ei]
```

with a live connection on the socket and `NRestarts=0` across a full session lifecycle.

Not tested: AMD hosts, touch/pen clients (no route, see above), and a client sending absolute-only pointer input for a long session (the relative-delta carry is exercised, but I have no absolute-only client to soak it with).

## Notes

No changes to pairing, trusted subnets, certificates, client commands, shell hooks, or privilege boundaries. Packaging: one new optional pkg-config dependency (`libei-1.0`), added to the Nix derivation here; other packaging flows would want the same.


papi-ux:
this is a lot of good work and the shape is right. the EIS server is the only way into a headless gamescope, XTEST is already a libei client so this is the same road, and the optional dependency is gated properly: with libei absent every one of these compiles to a stub that returns false, so nobody on uinput today changes. i want this in.

two of the problems are mine, not yours. no CI job compiles this code, theres no libei-dev in the build workflow and the only build that would define the feature is Build polaris-stream, which is dispatch and schedule only and shows SKIPPED on this PR. and only the nix package gained the dependency, so the rpm, arch, steamos and appimage builds all ship it compiled out, which means merged as it stands it doesnt reach anyone who isnt building from nix. ill fix both of those on my side.

three i need from you first.

1. `ensure_ready()` blocks. `pump(ctx && !emulating ? 250 : 0)` is on the path every move, button, scroll and key event takes, and polaris runs `task_pool.start(1)` (main.cpp:466). one thread, and it also carries key repeat, the click delay, the gamepad back button timeout and the 10 second force shutdown timers. any state that leaves the device paused makes every packet pay 250 ms on the thread thats supposed to be ending the session.
2. utf8 text still escapes to the host. `platf::keyboard::unicode()` only asks the wayland route and the new one has no `unicode()` at all, so moonlight text events still type ctrl+shift+u and hex digits into the host desktop. thats the thing your own comment says must not happen.
3. the gate is a config string rather than the running compositor. it checks `stream_mode == "gamescope_stream"` where our own predicate is that or `private_runtime == "gamescope"` (process.cpp:5107), and it never asks whether gamescope is actually up. since `should_block_host_fallback()` returns the same config only answer, a host set to gamescope whose compositor failed to start gets no input at all instead of falling back to uinput.

smaller things while youre in there:

- `move()` never re-anchors `last_absolute`, and moonlight mixes relative drag with absolute taps, so the first absolute event after any relative motion is a delta from a stale anchor and the offset sticks for the rest of the session
- when ctx is null every event rebuilds a sender and reconnects, no backoff, at motion rates
- one notch maps to value120 correctly, but the continuous axis value is 1.0 where libinput and our own labwc route use about 10, so a client reading `axis` instead of `axis_value120` scrolls about ten times too slow
- touch and pen go from reaching the host desktop to being dropped under gamescope_stream. probably right, but it needs to be in the release notes

is any of this something youd rather hand to me? happy to take the CI and packaging half either way.

kubicgruenfeld:
Thanks — and yes please, take the CI and packaging half. I have no way to test the rpm/arch/steamos/appimage paths, so those would be guesses from me. I've done the three code items; pushed as ba64cfca and 59d2afdf.

**1. `ensure_ready()` no longer blocks.** The handshake now runs on its own reader thread that polls the libei fd, so nothing on the event path waits at all — not 250 ms, not 1 ms. Consequences worth knowing:

- Only the *first* event of a session is dropped; the device is ready a millisecond or so later and every event after it finds it. It is dropped rather than leaked, since `owns_input()` still claims it.
- libei stays single-threaded-by-mutex: every `ei_*` call, on the reader thread and on the caller's thread alike, is made under `impl->mutex`.
- `join_pumper(bool force)` is the one sharp edge, so it carries a comment. The thread takes the mutex on every pass, so joining it from under the mutex would deadlock. `force` is only ever used from `shutdown()`, which releases the mutex first; the non-forcing call reaps a thread that has already run to completion, which it knows from an atomic the thread sets as its very last action, after its last unlock.
- Connection attempts are now a second apart, which covers your "no backoff at motion rates" point.

**2. `unicode()` added**, and `platf::keyboard::unicode()` now asks it and consults its `should_block_host_fallback()`. Same ibus hex entry the Wayland route types. Whether the focused client picks it up is its business; the point is that it stops being typed into the host desktop.

**3. The gate takes both signals now** — `stream_mode == "gamescope_stream" || private_runtime == "gamescope"` — and withdraws the claim once the socket has refused, so a host configured for gamescope whose compositor never started falls back to uinput.

While fixing that I found the same "no input at all" reachable by a second route: `owns_input()` has to keep withholding events while a connection is in flight, so a socket that *answers* and then stalls would withhold forever. 59d2afdf bounds it — a connection that has not offered a device within two seconds is dropped and marked failed, and a later attempt can still succeed.

Also done: `move()` drops the absolute anchor, since Moonlight mixes relative drag with absolute taps and a stale anchor made the offset stick for the session. And the touch/pen drop is written down in `docs/stream-paths.md`, under "Relation to gamescope", alongside why the Wayland route has nothing to bind here — happy to move it if you'd rather it lived somewhere else, and the release-note line would be something like *"Under Gamescope, touch and pen from the client are no longer forwarded to the host desktop."*

**One thing I did not change, and I think the review is wrong on this one — please check my reading.** The continuous axis is ~10 per notch in libinput and in our labwc route, but gamescope is not libinput here. `wlserver_mousewheel()` takes *notches*: it passes the value straight through as the continuous `wl_pointer` axis and derives `axis_value120` from that same value as `value * WLR_POINTER_AXIS_DISCRETE_STEP` (wlserver.cpp:2820). The two are locked at 1:120, so no client can get both. And gamescope's own physical-input paths feed it `value120 / 120.0` — `LibInputHandler.cpp:167` and `WaylandBackend.cpp:3127` both do exactly that. So 120 units per notch is what makes us scroll like the mouse plugged into the host; scaling to ~10 would fix legacy `axis` readers at the cost of making `axis_value120` report ten notches per notch, which is the number Steam actually reads. I left it at 120 with the reasoning in a comment. If you want the legacy readers served properly, I think that is a fix to gamescope's `wlserver_mousewheel()`, not to this route, and I'm glad to take it there.

**One note for the packaging half:** `POLARIS_BUILD_EI_VIRTUAL_INPUT` currently sits inside the `if(WAYLAND_FOUND)` block in `cmake/compile_definitions/linux.cmake`, so a build without Wayland dev headers drops libei even when libei is present. Probably wants lifting out when you wire up the other builders.

Verified compiling and linking on NixOS (libei 1.6.0, `libei.so.1` in `DT_NEEDED`, new paths present in the binary). Not yet re-tested on a live stream — I'll do that next and report back.

kubicgruenfeld:
Live-tested on NixOS now, gamescope_stream with a Steam session: mouse, keyboard and scroll all behave, and the dropped first event is not perceptible — no lost click at session start. So the reader thread covers the handshake well within the gap before the second event arrives, and pre-connecting at session start is not needed.

That was the only behavioural risk I flagged in the rework. From my side the three items are done and verified end to end.

papi-ux:
you were right about scroll and i was wrong. `wlserver_mousewheel()` takes notches and derives `axis_value120` from the same number, so 120 per notch is what makes it scroll like a mouse plugged into the host. leave it, and the comment you wrote is the right one.

i built your branch against libei 1.6 on fedora 44 with the real flags and it compiles clean, and the three items read right to me. two things i found while reading the reader thread, both small, so i fixed them on top of your branch instead of sending you around again:

- the reader polls the wake pipe but never reads it. `disconnect_locked()` writes a byte even when no reader is running, so a `reset()` after the reader was already reaped leaves one behind, and the next connection's reader sees the pipe readable on every poll and spins on the mutex. the reader drains it now.
- after the 2 s give up, the next event reconnects right away, the socket answers, `connect_failed` clears, and input is swallowed for another 2 s. so a socket that answers and then stalls gets two seconds swallowed, one event through, repeat. now the flag only clears once a device is actually emulating, a stalled socket is retried every 10 s, and the deadline only applies to a connection that never offered a device (a paused one was tripping it too).

there is a test for both against a listening socket nobody accepts on. three of four cases are red on your head (the reader burned 384 ms of a 400 ms window) and all four are green with the fixes.

the packaging half is done too: libei in the fedora spec, both cpack lists, both PKGBUILDs, the SteamOS build root (3.8 ships 1.4.1), the ubuntu, codeql, arch and opensuse CI, the docs, and the detection lifted out of `if(WAYLAND_FOUND)` like you said. every package job now checks that the binary it built actually needs `libei.so.1`, because configure skips it silently and a release could lose the route without anything failing.

i pushed those as two commits on top of yours since maintainer edits are on. if youd rather have them somewhere else just say so and ill move them. your release note line is good, ill use it as is.

kubicgruenfeld:
Thanks — and both of those are mine, both real.

The wake byte is the worse one. `drain_wake_pipe()` only ran in `join_pumper()`, so I was draining on the path where the reader had already stopped and never on the path where one was running. And since `disconnect_locked()` writes unconditionally, a `reset()` with no reader to reap leaves the byte sitting there for the next reader to find — which then wakes on every poll and takes the mutex the input path needs, which is the thing the whole reader-thread design exists to avoid. Draining it in the reader is the right place.

The `connect_failed` one I got backwards: I cleared it where the connection was made rather than where it started working, so every retry against a stalled socket handed the input back for another two seconds. Clearing it in `drain_locked()` on RESUMED is where it belongs — that is the first moment the route can actually carry anything. And you're right that `device_seen` is needed to keep the deadline off a device that was offered and later paused; I had conflated "never offered one" with "not emulating right now", and gamescope only ever calls `eis_device_resume`, so I never saw the difference locally.

Both commits on top of mine is exactly where I'd want them — no need to move anything. The tests are better than what I'd have written, particularly the one for the spin, which is the kind of thing that would otherwise only show up as someone reporting that input got choppy after a reconnect.

Noted on the packaging, especially making every package job assert the binary really needs `libei.so.1`. Silent configure skips were the failure mode I was most worried about and had no way to check.

I've rebased my own host onto master and it's running the merged version now. I'll report back if anything turns up in use.

PR #750, opened 2026-09-20
fix(clients): keep a profile's "auto" HDR state through a UI round trip

### LINKED ISSUE #749: Client profile HDR override: "auto" is serialised as false, so saving a profile silently forces HDR off
### Describe the bug

A client profile's HDR override is tri-state in the model:

```cpp
std::optional<bool> hdr;   ///< Override HDR enable/disable for this client
```

Unset means "no override — follow what the client asks for", which is the behaviour the web UI describes. But the API the UI reads collapses that to `false`:

```cpp
// get_all_profiles_json() — what the UI sees
entry["hdr"] = profile.hdr.has_value() ? profile.hdr.value() : false;

// save_profiles_to_disk() — what is actually persisted, tri-state intact
if (profile.hdr.has_value()) entry["hdr"] = profile.hdr.value();
```

The editor then flattens it again (`PinView.vue:1623`, `hdr: profile.hdr || false`) and renders a two-state checkbox (`PinView.vue:717`). So an unset profile is shown as an unticked box, and saving that form posts `"hdr": false`, which the loader reads back as an explicit **force off**:

```cpp
if (value.contains("hdr") && value["hdr"].is_boolean())
  profile.hdr = value["hdr"].get<bool>();
```

Net effect: **"auto" silently becomes "force off" as soon as the profile is saved from the UI**, and there is no way to get back to auto through the UI at all.

The write path can't accept the correction either — `body["hdr"].get<bool>()` (`confighttp.cpp:5281`) throws on `null`, so a client sending explicit null gets an error rather than a cleared override.

### Steps to reproduce

1. Pair a client and open its profile in the web UI. HDR appears unticked.
2. Save the profile without touching HDR.
3. `~/.config/polaris/client_profiles.json` now contains `"hdr": false`.
4. Start a stream **with** HDR enabled on the client. The host logs

```
client_profiles: Loaded profile for "<name>" … hdr=false
portal HDR force -> 0 from enable_hdr=false
HDR decision: client_dynamic_range=1 display_hdr=false … stream_hdr_enabled=false
Color coding: SDR (Rec. 709)   Color depth: 10-bit
```

The client negotiated a 10-bit HDR decode and the host sends SDR Rec.709 with no HDR metadata, so the picture arrives heavily oversaturated (strong red cast) — the client is applying an HDR transfer to SDR content.

Removing the `"hdr"` key by hand restores the documented behaviour: the profile loads as `hdr=(default)` and `enable_hdr` then tracks the client's request in both directions, confirmed on a device that requests HDR docked and SDR handheld.

### Install method

Local package build (Nix flake from the repo's own `nix/` module).

### Host OS

NixOS 26.11

### GPU and driver

NVIDIA RTX 4070 Ti Super, driver 610.57.04

### Runtime path

Headless (gamescope_stream)

### Expected behavior

An unset HDR override should stay unset through a UI round trip, and the UI should be able to express all three states — the point of the tri-state is exactly the case of a device that wants HDR sometimes and not others (a Steam Deck docked to an HDR TV versus handheld).

Concretely: emit `null` for unset rather than `false`, accept `null` on the way in as "clear the override", and offer Auto / Enabled / Disabled instead of a checkbox.

I have a patch and will open it as a PR.

### Additional context

`color_range` is serialised with the same `has_value() ? … : 0` shape in `get_all_profiles_json()`. I haven't checked whether `0` is a meaningful value there or also stands in for "unset", so I've left it alone — worth a look by someone who knows the intent.

kubicgruenfeld:
## Summary

Fixes #749.

`client_profile_t::hdr` is `std::optional<bool>` — unset means "no override, follow what the client asks for". Three places flattened that to `false`:

```cpp
// get_all_profiles_json() — what the UI reads
entry["hdr"] = profile.hdr.has_value() ? profile.hdr.value() : false;
```
```js
// PinView.vue, loading a profile into the editor
hdr: profile.hdr || false,
// …and a two-state checkbox to render it
```

So an unset profile showed as an unticked box, and saving the form posted `"hdr": false`, which loads back as an explicit **force off**. Auto became force-off just by saving a profile, and there was no way to restore it from the UI.

That is worse than a stuck setting. With a profile forced off and a client that asks for HDR, the host builds an SDR session while the client negotiated a 10-bit HDR decode — SDR Rec.709 arrives where the client expects PQ, and the picture comes out heavily oversaturated:

```
client_profiles: Loaded profile for "<name>" … hdr=false
portal HDR force -> 0 from enable_hdr=false
HDR decision: client_dynamic_range=1 display_hdr=false … stream_hdr_enabled=false
Color coding: SDR (Rec. 709)   Color depth: 10-bit
```

This changes three things:

1. `get_all_profiles_json()` emits `null` for unset instead of `false`.
2. `updateClientProfile` treats `null` as "clear the override" — `get<bool>()` would have thrown on it, so a client could not previously send the correction either.
3. The editor offers **Auto / On / Off** instead of a checkbox, and loads with `profile.hdr ?? null` so unset stays distinct from off.

Auto is what a device wanting HDR only sometimes needs — a Steam Deck docked to an HDR TV versus handheld — and it is what the tri-state was already modelling; only the UI contract was missing.

The presentation is the opinionated part: a `<select>` seemed the plainest way to express three states in the existing form, but I'm happy to change it if you'd rather have something else.

## Testing

Tested on NixOS 26.11, RTX 4070 Ti Super, driver 610.57.04, `gamescope_stream`, with two paired clients.

The behaviour this restores was verified end to end by clearing the key by hand on a running host: profiles then load as `hdr=(default)` and `enable_hdr` tracks the client's request in both directions, confirmed on a device that asks for HDR docked and SDR handheld. The oversaturated picture goes away with it.

Being straight about coverage: the backend round trip (null out, null in, key omitted on disk, unset on reload) I have traced end to end, and `en.json` parses. The Vue change I have **not** exercised in a browser — no build of the web UI was run here. Worth a look by someone who can click through it before merging.

I did leave two pre-existing `Boolean(profile.hdr)` uses alone (the AI-suggestion match and the profile summary label); they treat auto as off, which is the same as today's behaviour and seemed out of scope.

## Notes

`color_range` is serialised with the same `has_value() ? … : 0` shape in `get_all_profiles_json()`. I did not touch it because I don't know whether `0` is a meaningful value there or also stands in for unset — flagged in #749 for someone who knows the intent.

No changes to pairing, trusted subnets, certificates, client commands, shell hooks, packaging scripts, or privilege boundaries. One new user-facing string (`pin.hdr_auto`) in `en.json` only; other locales fall back.


papi-ux:
good catch, and the round trip was the right way to show it. the save path already omitted `hdr` when unset and the loader ignores anything that isnt a boolean, so the file on disk was always right and it was only the api and the checkbox flattening it. merging, thank you!

